PR #6, opened 2024-03-06
Upgrade SDK base image to Ubuntu 23.10

nikolaszimmermann:
Didn't do extensive testing yet - but things just appear to work fine.


nikolaszimmermann:
Let's close this, and move straight to 24.04. We do need clang18 now in the SDK, otherwise the Linux/clang build is broken. Also important if we want to switch the bots to the new SDK.

PR #16, opened 2024-04-08
Attempt to build wkdev-sdk for multiple target platforms

nikolaszimmermann:
Still WIP.

### DIFF
--- a/.github/workflows/wkdev-sdk.yml
+++ b/.github/workflows/wkdev-sdk.yml
@@ -37,8 +37,8 @@ jobs:
           remove-codeql: true
           remove-docker-images: true
 
-      - name: Install podman
-        run: sudo apt-get update && sudo apt-get -y install podman fuse-overlayfs
+      - name: Install dependencies
+        run: sudo apt-get update && sudo apt-get -y install podman fuse-overlayfs qemu-user-static
 
       - name: Checkout repo
         uses: actions/checkout@v4
@@ -46,7 +46,7 @@ jobs:
       - name: Build image
         run: |
           source ./register-sdk-on-host.sh
-          wkdev-sdk-bakery --mode=build --verbose
+          wkdev-sdk-bakery --mode=build --platforms=linux/arm/v7,linux/amd64 --verbose
           wkdev-sdk-bakery --mode=export --verbose
 
       - name: Archive image
--- a/scripts/host-only/wkdev-sdk-bakery
+++ b/scripts/host-only/wkdev-sdk-bakery
@@ -19,6 +19,7 @@ argsparse_use_option =verbose     "Increase verbosity of this script"
 
 argsparse_use_option =name:       "Name of container image" mandatory default:wkdev-sdk
 argsparse_use_option =mode:       "Operation mode: 'build', 'deploy', or 'export'" mandatory
+argsparse_use_option =platforms:  "Build container for given set of platforms" mandatory default:"linux/amd64"
 argsparse_use_option idle-cores:  "Number of CPU cores to leave idle, when building the image" type:uint default:2
 
 argsparse_usage_description="$(cat <<EOF
@@ -45,6 +46,7 @@ process_command_line_arguments() {
     idle_cores=${program_options["idle-cores"]}
 
     operation_mode="${program_options["mode"]}"
+    platforms="${program_options["platforms"]}"
     [ "${operation_mode}" != 'build' ] && [ "${operation_mode}" != 'deploy' ] && [ "${operation_mode}" != 'export' ] && _abort_ "Unknown operation mode: '${operation_mode}' (valid choices are 'build', 'deploy', or 'export')"
 }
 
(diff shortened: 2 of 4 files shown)


TingPing:
FWIW I was just waiting for Github Actions to have public ARM runners.

Otherwise I've been using it on my ARM laptop already.

nikolaszimmermann:
> FWIW I was just waiting for Github Actions to have public ARM runners.
I've been toying locally with qemu-user-static + podman build --platform support, and it works just fine out of the box.
We  need a way to produce e.g. armv7 32bit images or risc-v, etc. so I was looking into the qemu solution, as a general thing. Since we don't rebuild images so often I think increased build times are not a concern.

> Otherwise I've been using it on my ARM laptop already.
Great to hear!

scripts/host-only/wkdev-sdk-bakery
```diff
@@ -19,6 +19,7 @@ argsparse_use_option =verbose     "Increase verbosity of this script"
 
 argsparse_use_option =name:       "Name of container image" mandatory default:wkdev-sdk
 argsparse_use_option =mode:       "Operation mode: 'build', 'deploy', or 'export'" mandatory
+argsparse_use_option =platforms:  "Build container for given set of platforms" mandatory default:"linux/arm/v7,linux/amd64"
```
TingPing, line 22: I'd think the default should just be current architecture, with CI doing more.

TingPing:
> I've been toying locally with qemu-user-static + podman build --platform support, and it works just fine out of the box.

Sounds great!

scripts/host-only/wkdev-sdk-bakery
nikolaszimmermann: Correct.

.github/workflows/wkdev-sdk.yml
```diff
@@ -36,7 +36,10 @@ jobs:
           remove-haskell: true
           remove-codeql: true
           remove-docker-images: true
-      
+
+      - name: Install qemu-user-static
+        run: sudo apt-get -y install qemu-user-static
+
```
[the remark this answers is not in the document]
nikolaszimmermann: Ok.

.github/workflows/wkdev-sdk.yml
```diff
@@ -36,20 +36,20 @@ jobs:
...
+        run: sudo apt-get -y install fuse-overlayfs qemu-user-static
 
       - name: Checkout repo
         uses: actions/checkout@v4
 
       - name: Build image
         run: |
           source ./register-sdk-on-host.sh
-          wkdev-sdk-bakery --mode=build --verbose
+          wkdev-sdk-bakery --mode=build --platforms=linux/arm/v7,linux/amd64 --verbose
```
TingPing, line 52: Could we actually make this build both in parallel?

https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

scripts/host-only/wkdev-sdk-bakery
```diff
@@ -69,7 +73,8 @@ deploy_image() {
 
     local image_directory="$(get_image_directory_by_name "${container_image_name}")"
     pushd "${image_directory}" &>/dev/null || _abort_ "Switching to directory '${image_directory}' failed"
-    run_podman_silent_unless_verbose push "$(get_tag_for_build)" || _abort_ "Pushing to registry failed"
+    run_podman_silent_unless_verbose push "$(get_tag_for_build)" || _abort_ "Pushing image to registry failed"
```
TingPing, line 76: I think you have to modify `get_tag_for_build` to include arch information. I don't think two images can have the same tag.

.github/workflows/wkdev-sdk.yml
nikolaszimmermann: Will try that, great suggestion.

scripts/host-only/wkdev-sdk-bakery
nikolaszimmermann: Hmm, I thought that is not necessary since the information is in the manifests?

nikolaszimmermann:
```
$ podman manifest inspect ghcr.io/igalia/wkdev-sdk:multi-arch
{
    "schemaVersion": 2,
    "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
    "manifests": [
        {
            "mediaType": "application/vnd.oci.image.manifest.v1+json",
            "size": 2730,
            "digest": "sha256:fb24d36b418730f31d33693a9e3e077e15ad3421879789981b22fd700cfcbb23",
            "platform": {
                "architecture": "amd64",
                "os": "linux"
            }
        },
        {
            "mediaType": "application/vnd.oci.image.manifest.v1+json",
            "size": 2730,
            "digest": "sha256:084dff9e22c8f875278e0f55ce1085a0c7e9952c8a2f58662b4408038c9f4bed",
            "platform": {
                "architecture": "arm",
                "os": "linux",
                "variant": "v7"
            }
        }
    ]
}
```

.github/workflows/wkdev-sdk.yml
nikolaszimmermann: That requires a rework of the wkdev-sdk-bakery if we want that.
We need to move 'image manifest' creation out of wkdev-sdk-bakery, execute that first, then build image for platform A, then for platform B, and each time register the image within the manifest. Then finally push the manifest, together with the images, building a single multi-arch container including both architectures, under the same name/tag.

IMHO I'd leave it as-is, until we really see the image build time as an issue.

TingPing (approved):
We still need the build to consistently succeed, but all sounds good to me.

TingPing:
I believe this next build will pass :crossed_fingers:

But the fact this goes from 40 minutes to 3-4 hours sucks... I'd like to move it to a matrix to parallelize it. It is possible to create a manifest that points to other images so it won't appear any different even if deployed separate.

PR #28, opened 2024-06-28
Document how to use cross-toolchain-helper with the SDK

nikolaszimmermann:


alexgcastro (approved):
LGTM! Very useful. I would say maybe that we recommend to do it from inside because we usually test it in that distro.

nikolaszimmermann:
I have no strong opinion, but it should definitely work out of the box in the container. I would say we hold back our recommendation until the meson setup issue is understood / fixed. @clopez already said he's going to have a look next week - it is also affecting non-container builds by the way.

PR #29, opened 2024-06-28
Build our own sccache v0.8.1 package.

nikolaszimmermann:
The system provided Ubuntu v0.7.7 package is broken:

```
📦 nzimmermann@wkdev:~/$ cat foo.cpp
int foo() {{
📦 nzimmermann@wkdev:~/$ /usr/bin/sccache g++ -o foo.o -c foo.cpp foo.cpp: In function 'int foo()':
foo.cpp:1:13: error: expected '}' at end of input
    1 | # 0 "foo.cpp"
      |            ~^
foo.cpp:1:13: error: expected '}' at end of input
    1 | # 0 "foo.cpp"
      |           ~ ^
foo.cpp:1:13: warning: no return statement in function returning non-void [-Wreturn-type]
sccache: Compiler killed by signal 1
📦 nzimmermann@wkdev:~/$ /usr/bin/sccache g++ -o foo.o -c foo.cpp
📦 nzimmermann@wkdev:~/$
```

### DIFF
--- a/images/wkdev_sdk/Containerfile
+++ b/images/wkdev_sdk/Containerfile
@@ -127,5 +127,8 @@ RUN export QT_VERSION=$(qmake6 -query QT_VERSION) && \
       ln -s ${directory} ${directory}/${QT_VERSION}  >/dev/null 2>&1 || true; \
     done
 
+# Build custom sccache version (the Ubuntu provided 0.7.7 package is broken)
+RUN cargo install --root /usr/local --version 0.8.1 --locked sccache
+
 # Switch back to interactive prompt, when using apt.
 ENV DEBIAN_FRONTEND dialog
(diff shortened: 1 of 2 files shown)


images/wkdev_sdk/Containerfile
```diff
@@ -127,5 +127,8 @@ RUN export QT_VERSION=$(qmake6 -query QT_VERSION) && \
       ln -s ${directory} ${directory}/${QT_VERSION}  >/dev/null 2>&1 || true; \
     done
 
+# Build custom sccache version (the Ubuntu provided 0.7.7 package is broken)
+RUN cargo install --root /usr --version 0.8.1 --locked sccache
```
TingPing, line 131: I might prefer `/usr/local` but pleasantly easy to build. :+1:

TingPing (approved):

images/wkdev_sdk/Containerfile
nikolaszimmermann: Indeed, thinking about it /usr isist is for system packages nothing self-built. I will adapt the PR.

PR #33, opened 2024-07-31
Export PIP_BREAK_SYSTEM_PACKAGES=1 by default.

nikolaszimmermann:
This avoids the error message when installing python packages, that on Ubuntu you should install the python-xxxx packages instead of using pip3 install, if possible.

When that is set the error is suppressed, and one doesn't need to pass '--break-system-packages' to python manually.


alexgcastro (approved):
LGTM!

PR #34, opened 2024-08-01
scripts: Mount /run/udev for gamepads

cadubentzen:
To use an Xbox controller via libmanette, `/run/udev` needs to be mounted. Otherwise I see errors such as `openat(AT_FDCWD<..>, "/run/udev/data/c13:63", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)` and the gamepad isn't detected.


nikolaszimmermann:
Looks good, thanks!

PR #36, opened 2024-08-19
List libsysprof-capture-4-dev for installation

aperezdc:
WebKit uses libsysprof-capture by default to gather runtime profiling data. Currently developer builds use a bundled copy of the library, so provide it in the SDK container to allow switching to always preferring the system-installed copy.

----

This is needed before we can tackle https://bugs.webkit.org/show_bug.cgi?id=277627

### DIFF
--- a/images/wkdev_sdk/required_system_packages/04-devtools.lst
+++ b/images/wkdev_sdk/required_system_packages/04-devtools.lst
@@ -8,7 +8,8 @@ icecc ccache cargo
 valgrind rr perf-tools-unstable systemd-coredump
 
 # Documentation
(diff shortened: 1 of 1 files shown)


aperezdc:
As @TingPing pointed out, `images/wkdev_sdk/Containerfile` calls into `Tools/{wpe,gtk}/install-dependencies` into the WebKit checkout, so adding the package to the lists used for those scripts would have it installed without needing this patch. Closing, and I will make the corresponding patch for WebKit instead.

TingPing:
Triggered a new build so it will be added to the image: https://github.com/Igalia/webkit-container-sdk/actions/runs/10202240105

PR #42, opened 2024-09-06
GitHub Actions: Free up more space in CI worker

dpino:
The current `wkdev-sdk.yml` contains an action that frees up space in the GitHub Action CI worker. In addition to that, I want to try out another script that removes even more unnecessary stuff. If the gain is significant, we can keep it.

### DIFF
--- a/.github/workflows/scripts/free_disk_space.sh
+++ b/.github/workflows/scripts/free_disk_space.sh
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+#
+# Based on: https://github.com/apache/flink/blob/master/tools/azure-pipelines/free_disk_space.sh
+#
+echo "=============================================================================="
+echo "Freeing up disk space on CI system"
+echo "=============================================================================="
+echo ""
+
+# Before
+echo "Disk space before:"
+df -h
+echo ""
+
+# List packages
+echo "Listing 100 largest packages"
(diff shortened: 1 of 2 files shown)


dpino:
It doesn't seem it's freeing up a lot of space (1GB Max) https://github.com/Igalia/webkit-container-sdk/actions/runs/10734917079/job/29771089547?pr=42

Perhaps the script can be tweaked to remove more unncessary things.

.github/workflows/scripts/free_disk_space.sh
```diff
@@ -0,0 +1,60 @@
...
+sudo apt-get clean
+echo ""
+echo "Disk space after apt-get:"
+df -h
+echo ""
+
+# Large dirs
+echo "Removing large directories"
+# https://github.com/apache/flink/blob/master/tools/azure-pipelines/free_disk_space.sh
+sudo rm -rf /usr/share/dotnet/
```
TingPing, line 39:
```suggestion
```

We already do this.

TingPing (approved):
Seems fine.

.github/workflows/scripts/free_disk_space.sh
dpino: Right, besides dotnet the GitHub Action that runs before this script was also removing Android and Haskell so I removed those steps from the script.

PR #46, opened 2024-09-16
jhbuild: Backport webrtc patches from GStreamer main branch

philn:
Those will not ship until 1.26.

https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/7156 https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/7294

### DIFF
--- a/images/wkdev_sdk/jhbuild/webkit-sdk-deps.modules
+++ b/images/wkdev_sdk/jhbuild/webkit-sdk-deps.modules
@@ -108,7 +108,11 @@
     <branch repo="gstreamer.freedesktop.org"
             checkoutdir="gstreamer"
             module="gstreamer.git"
-            tag="1.24.8"/>
+            tag="1.24.8">
+      <patch file="0001-webrtcbin-create-and-associate-transceivers-earlier-.patch" strip="1"/>
+      <patch file="0002-webrtcbin-reverse-direction-from-remote-media.patch" strip="1"/>
+      <patch file="0003-webrtcbin-connect-output-stream-on-recv-transceivers.patch" strip="1"/>
+    </branch>
     <dependencies>
       <dep package="openh264"/>
     </dependencies>
--- a/images/wkdev_sdk/Containerfile
+++ b/images/wkdev_sdk/Containerfile
@@ -93,6 +93,7 @@ RUN rustup default stable && \
 # Copy jhbuild helper files and do the initial build & install
 COPY /jhbuild/jhbuildrc /etc/xdg/jhbuildrc
 COPY /jhbuild/webkit-sdk-deps.modules /jhbuild/webkit-sdk-deps.modules
+COPY /jhbuild/patches/ /jhbuild/patches/
 WORKDIR /jhbuild
 RUN git clone https://gitlab.gnome.org/GNOME/jhbuild.git && \
     cd jhbuild && \
--- a/images/wkdev_sdk/jhbuild/patches/0001-webrtcbin-create-and-associate-transceivers-earlier-.patch
+++ b/images/wkdev_sdk/jhbuild/patches/0001-webrtcbin-create-and-associate-transceivers-earlier-.patch
@@ -1,0 +1,912 @@
+From 48ae40f477523bed4cd709d163f541e748356071 Mon Sep 17 00:00:00 2001
+From: Carlos Bentzen <cadubentzen@igalia.com>
+Date: Wed, 10 Jul 2024 10:34:19 +0200
+Subject: [PATCH] webrtcbin: create and associate transceivers earlier in
+ negotation
+
+According to https://w3c.github.io/webrtc-pc/#set-the-session-description
+(steps in 4.6.10.), we should be creating and associating transceivers when
+setting session descriptions.
+
+Before this commit, webrtcbin deviated from the spec:
+1. Transceivers from sink pads where created when the sink pad was
+   requested, but not associated after setting local description, only
+   when signaling is STABLE.
+2. Transceivers from remote offers were not created after applying the
+   the remote description, only when the answer is created, and were then
+   only associated once signaling is STABLE.
+
+This commit makes webrtcbin follow the spec more closely with regards to
+timing of transceivers creation and association.
+
+A unit test is added, checking that the transceivers are created and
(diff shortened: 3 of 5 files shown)


images/wkdev_sdk/jhbuild/webkit-sdk-deps.modules
```diff
@@ -99,7 +99,11 @@
     <branch repo="gstreamer.freedesktop.org"
             checkoutdir="gstreamer"
             module="gstreamer.git"
-            tag="1.24.6"/>
+            tag="1.24.6">
+      <patch file="0001-webrtcbin-create-and-associate-transceivers-earlier-.patch" strip="1"/>
+      <patch file="0002-webrtcbin-reverse-direction-from-remote-media.patch" strip="1"/>
+      <patch file="0003-webrtcbin-connect-output-stream-on-recv-transceivers.patch" strip="1"/>
```
TingPing, line 105: Patches don't apply to git sources. So this would have to either use a tarball or change the commit to one that includes these.

philn: The former then, because the latter is not possible (we shouldn't use git main imho). Thanks for the review! I'll update the PR tomorrow.

philn: Oh but then we'd need to split modules again (-base, -good, -bad, etc), huh... sigh :(

philn: > Patches don't apply to git sources

Whyyyy?

TingPing: Its unfortunate, I'm not sure the history behind it. One day I'd like to fix that in jhbuild.

nikolaszimmermann:
Thanks @philn ! In the meanwhile we upgraded to gst 1.26. -- are those patches still relevant?

philn:
You upgraded to 1.24.6 which doesn't ship these patches.

nikolaszimmermann:
> You upgraded to 1.24.8 which doesn't ship these patches.

Oops - misremembered 😂 thanks for the clarification.

images/wkdev_sdk/jhbuild/webkit-sdk-deps.modules
dpino: https://gitlab.gnome.org/GNOME/jhbuild/-/merge_requests/257 r?

TingPing: This was merged so this can be rebuilt and tested.

philn:
Error: identifier is not a container: preparing container for next step: creating build container: creating container: creating read-write layer with ID "4244d8e8bb5a3d50479680d92c5091a3b8f9d9138a512a124a9f0de95a44c90e": no space left on device

dpino:
> Error: identifier is not a container: preparing container for next step: creating build container: creating container: creating read-write layer with ID "4244d8e8bb5a3d50479680d92c5091a3b8f9d9138a512a124a9f0de95a44c90e": no space left on device

I will take a look.

dpino:
@philn I freed up space in the host. Now it's failing with a different error:

```
*** Installing sysprof *** [13/15]
meson install --no-rebuild --quiet
/bin/sh /usr/local/share/jhbuild/triggers/desktop-database.trigger
/bin/sh /usr/local/share/jhbuild/triggers/gtk+.trigger
/bin/sh /usr/local/share/jhbuild/triggers/mime-database.trigger
*** module libspiel not built due to non buildable gstreamer *** [14/15]
*** module webkit-sdk-deps not built due to non buildable libwpe *** [15/15]
*** module webkit-sdk-deps not built due to non buildable wpebackend-fdo *** [15/15]
*** module webkit-sdk-deps not built due to non buildable gstreamer *** [15/15]
*** module webkit-sdk-deps not built due to non buildable libspiel *** [15/15]
*** the following modules were not built *** [15/15]
libwpe wpebackend-fdo gstreamer libspiel webkit-sdk-deps 
Error: building at STEP "RUN git clone https://gitlab.gnome.org/GNOME/jhbuild.git &&     cd jhbuild &&     ./autogen.sh --prefix=/usr/local &&     make &&     make install &&     export JHBUILD_RUN_AS_ROOT=1 WKDEV_IN_IMAGE_BUILD=1 &&     jhbuild --no-interact build &&     rm -r /var/tmp/jhbuild/build": while running runtime: exit status 1

[FATAL] Container image build failed - aborting with exit code 1.
Error: Process completed with exit code 1.
```

https://github.com/Igalia/webkit-container-sdk/actions/runs/11381867202/job/31672580197?pr=46

philn:
Looks like a new jhbuild bug.

philn:
Related to disk usage, the gst monorepo has a "medias" subproject which is quite big. For our use-case that subproject is useless, can it be ignored during the git clone?

dpino:
> Looks like a new jhbuild bug.

Indeed, I was hiting the same error on Playwright. I pushed a patch to fix it:

- [cmake: Do not skip configure phase if CMakeCache.txt doesn't exist](https://gitlab.gnome.org/GNOME/jhbuild/-/merge_requests/266)

What's happening is that recently a patch to skip the `configuration` step for `cmake` modules was merged. The patch introduced a regression, and now all freshly pulled cmake modules skip the `configuration` step. Without running the `configuration` step is not possible to `build`.

The error you're seeing:
```
Cannot build: libwpe wpebackend-fdo gstreamer libspiel webkit-sdk-deps
```
happens because `libwpe` is a `cmake` module, which failed to build due to this regression. Everything after fails because there's a chain of dependencies: `wpebackend-fdo` depends on `libwpe`, `gstreamer` depends on `wpebackend-fdo` and `libspiel` depends on `gstreamer`.

Also, there's another bug in JHBuild that have been around for a long time and it's quite annoying. Whenever a module fails due to an error, there's not error message printed out in console. That makes it hard to figure out what has happened when a module fails. I also pushed a patch to fix this issue:

- [Print out error message when exiting on error](https://gitlab.gnome.org/GNOME/jhbuild/-/merge_requests/265)

cc @TingPing

PR #51, opened 2024-09-27
Upgrade jhbuild dependencies

nikolaszimmermann:
Switch to new libwpe/wpebackend-fdo releases, upgrade libdex/libpanel/etc.
Make sure to always attach tags to checkouts, to guarantee a certain version is used, independant of the image build time (where possible).

This includes the PR #50, and needs to rebased, once it's merged to main.

### DIFF
--- a/images/wkdev_sdk/jhbuild/webkit-sdk-deps.modules
+++ b/images/wkdev_sdk/jhbuild/webkit-sdk-deps.modules
@@ -50,8 +50,8 @@
     <branch repo="wpewebkit.org"
             checkoutdir="libwpe"
             module="libwpe-${version}.tar.xz"
-            version="1.14.2"
-            hash="sha256:8ae38022c50cb340c96fdbee1217f1e46ab57fbc1c8ba98142565abbedbe22ef"/>
+            version="1.16.0"
+            hash="sha256:c7f3a3c6b3d006790d486dc7cceda2b6d2e329de07f33bc47dfc53f00f334b2a"/>
   </cmake>
 
   <meson id="wpebackend-fdo">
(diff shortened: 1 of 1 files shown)


dpino:
Looks good to me. A nitpick would be to put the `tag` attribute in a new line, since it's in a newline in some modules and it's the same line as attribute `module` in others.

A follow-up would be to add the `version` number to all modules and the corresponding `pkg-config` filename. Whe the `pkg-config` and `version` numbers are present, JHBuild determines whether the module is already provided by system libraries, and if that's the case the module is not built. If either the `version` or `pkg-config` are not present, JHBuild always builds the module. So, for instance in the case of the module `openh264`:

```xml
   <meson id="openh264" mesonargs="-Dtests=disabled">
    <branch repo="github.com"
            checkoutdir="openh264"
            module="cisco/openh264.git"
            tag="v2.4.1" >
    </branch>
  </meson>
```

Ubuntu 24.04 features this version of openh264:

```bash
$ apt-cache madison libopenh264-dev
libopenh264-dev | 2.4.1+dfsg-1 | http://mirrors.kernel.org/ubuntu noble/universe amd64 Packages
```

If `version` and `pkg-config` were present in the module configuration, JHBuild would compare the version stated in `pkg-config` with the required `version` of the JHBuild module. And if both versions match (or pkg-config's version is superior), the module is skipped. The configuration would be:

```xml
   <meson id="openh264" mesonargs="-Dtests=disabled">
    <pkg-config>openh264.pc</pkg-config>
    <branch repo="github.com"
            checkoutdir="openh264"
            module="cisco/openh264.git"
            tag="v2.4.1" 
            version="2.4.1">
    </branch>
  </meson>
```

The pkg-config file is usually under `/usr/`:

```bash
$ find /usr -name "*.pc" | grep -m1 264
/usr/lib/x86_64-linux-gnu/pkgconfig/openh264.pc
```
```bash
$ grep -i version /usr/lib/x86_64-linux-gnu/pkgconfig/openh264.pc
Version: 2.4.1
```

So, since openh264's pkg-config version matches the required module version, the module is skipped.

Anyway, I think it's better to do this in a separated pull-request.

dpino (approved):

PR #54, opened 2024-09-27
[CI] Remove actions that freed up space in hosted GitHub runner

dpino:
These actions are not needed in the self-hosted runner.

Also, assign a temporal name to the wkdev container in the 'Test image' step. Assigning a temporal name to the container prevents potential name collisions.


nikolaszimmermann (approved):
Let's try it

.github/workflows/wkdev-sdk.yml
```diff
@@ -41,11 +41,13 @@ jobs:
...
           source ./register-sdk-on-host.sh
-          wkdev-create --create-home --verbose --attach --no-pull
-          wkdev-enter --exec -- git clone --depth=1 https://github.com/WebKit/WebKit.git
-          wkdev-enter --exec -- ./WebKit/Tools/Scripts/build-webkit --wpe --release --generate-project-only
-          wkdev-enter --exec -- ./WebKit/Tools/Scripts/build-webkit --gtk --release --generate-project-only
+          wkdev-create --create-home --verbose --attach --no-pull -n ${CONTAINER}
+          wkdev-enter -n ${CONTAINER} --exec -- git clone --depth=1 https://github.com/WebKit/WebKit.git
+          wkdev-enter -n ${CONTAINER} --exec -- ./WebKit/Tools/Scripts/build-webkit --wpe --release --generate-project-only
+          wkdev-enter -n ${CONTAINER} --exec -- ./WebKit/Tools/Scripts/build-webkit --gtk --release --generate-project-only
+          podman stop ${CONTAINER} || true
```
nikolaszimmermann, line 50: Good catch, that was indeed missing.
But we also want` podman rm ${CONTAINER} here`.

I didn't want to hold back merging for that, we can fix that in a follow-up pr.

Issue #49, opened 2024-09-26
wkdev-init failure regarding /dev/log


nikolaszimmermann:
On my machine I always see this error message:

```
[9/14] Set /dev/log symbolic link to /run/systemd/journal/dev-log ...
ln: failed to create symbolic link '/dev/log': Permission denied
```

wkdev-init proceeeds without problems, but we should fix that at some point, or understand what is happening.

nikolaszimmermann:
@psaavedra - you added support for this back then, any idea what we could do here regarding /dev/log?

psaavedra:
Related commit: 3a4c529da4b6e6ed7ccad1118353abafcacae5f9.

```
commit 3a4c529da4b6e6ed7ccad1118353abafcacae5f9
Author: Pablo Saavedra <psaavedra@igalia.com>
Date:   Tue Sep 26 12:30:54 2023 +0200

    wkdev-init: Set /dev/log symbolic link to /run/systemd/journal/dev-log
    
    For compatibility reasons creates the symlink to /dev/log for programs
    that expect the traditional syslog logging interface. Example: logger.

```

psaavedra:
Is /dev read-only in the latest container?

psaavedra:
It looks like that symbolic link is now already created in the system: `lrwxrwxrwx 1 nobody nogroup 28 Jul  7 09:40 /dev/log -> /run/systemd/journal/dev-log` so the  
commit 3a4c529da4b6e6ed7ccad1118353abafcacae5f9 is not longer needed for the latest ubuntu image. 

We can do 2 things: 
* Just remove/revert this change
* Improve the action and create the symbolic link just in case this doesn't exist.

What could be the preference choice?

psaavedra:
First suggestion for fixing the issue: https://github.com/Igalia/webkit-container-sdk/pull/58

nikolaszimmermann:
I think you’re proposed solution is fine, always guarantees /dev/log is available.

PR #58, opened 2024-09-29
Improve /dev/log symbolic link setup to only create it if missing (#49)

psaavedra:
This fixes cases where the symbolic link is already present in the system, such as in the latest Ubuntu images.

If `/dev/log` already exists, the script skips the link creation to avoid the following error message:

```
[9/14] Set /dev/log symbolic link to /run/systemd/journal/dev-log ...
ln: failed to create symbolic link '/dev/log': Permission denied
```

Issue: https://github.com/Igalia/webkit-container-sdk/issues/49

### DIFF
--- a/scripts/container-only/.wkdev-init
+++ b/scripts/container-only/.wkdev-init
@@ -133,8 +133,15 @@ try_setup_setgid_subuid_files() {
 
 try_setup_journal_dev_log() {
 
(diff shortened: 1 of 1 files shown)


nikolaszimmermann (approved):
Thanks, Pablo, looks great

PR #59, opened 2024-09-30
Pass argument '--rm' to avoid wkdev-create container name collission

dpino:
Follow-up to d347129.

Command `wkdev-create` allows passing argument `--rm` which stops and removes a container in case it already exists. This can be useful to avoid the container name collission, instead of creating a temporal name for the container and stopping and removing it late.

To avoid the home directory collission, `wkdev-create` can delete the home directory in case it already exists and `create-home` was passed as an argument.

### DIFF
--- a/scripts/host-only/wkdev-create
+++ b/scripts/host-only/wkdev-create
@@ -85,6 +85,10 @@ process_command_line_arguments() {
     host_hostname="$(hostnamectl hostname)"
     container_hostname="$(basename "${container_name}").${host_hostname}"
 
+    if argsparse_is_option_set "create-home" && -d "${container_user_home}"; then
+        rm -rf "${container_user_home}"
+    fi
+
     # Verify the container home directory is accessible.
     if [ ! -d "${container_user_home}" ]; then
         if ! argsparse_is_option_set "create-home"; then
--- a/.github/workflows/wkdev-sdk.yml
+++ b/.github/workflows/wkdev-sdk.yml
@@ -41,15 +41,11 @@ jobs:
 
       - name: Test image
         run: |
-          CONTAINER="wkdev-$(date +%s)"
           source ./register-sdk-on-host.sh
-          wkdev-create --create-home --home ${HOME}/${CONTAINER}-home --verbose --attach --no-pull --name ${CONTAINER}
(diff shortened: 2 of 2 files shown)


dpino:
I'm not sure about this change. What happens if two processes are running this step simultaneously? Giving an unique name to each container guarantees isolation. Perhaps what it makes sense is to have a `wkdev-remove` command to remove a container (and home directory) in a clean way.

TingPing:
I think switching to a tmp dir instead of based on date sounds fine.

scripts/host-only/wkdev-create
```diff
@@ -85,6 +85,10 @@ process_command_line_arguments() {
     host_hostname="$(hostnamectl hostname)"
     container_hostname="$(basename "${container_name}").${host_hostname}"
 
+    if argsparse_is_option_set "create-home" && -d "${container_user_home}"; then
+        rm -rf "${container_user_home}"
```
nikolaszimmermann, line 89: I was always afraid about adding an option that does an rm -rf :-)

dpino: I think we need a counterpart of `wkdev-create` but to remove a container.

Now in the runner there's:

```bash
$ podman ps
CONTAINER ID  IMAGE                            COMMAND               CREATED      STATUS          PORTS       NAMES
97f83258b59f  ghcr.io/igalia/wkdev-sdk:latest  /wkdev-sdk/script...  3 weeks ago  Up 3 weeks ago              wkdev-1727468367
fa1c67dc62f4  ghcr.io/igalia/wkdev-sdk:latest  /wkdev-sdk/script...  3 weeks ago  Up 4 days ago               wkdev
```

For some reason, the temporal container `wkdev-1727468367` was not removed.

nikolaszimmermann: We usually used podman rm directly. However, I'm not objecting to a wkdev-destroy script.

ntrrgc: I would be pretty upset if `wkdev-create` deleted my WebKit SDK home directory. I've always considered it permanent. That's where you place your `.wkdev-firstrun`, after all.

nikolaszimmermann:
I think it's safe to resolve this as WONTFIX.

PR #64, opened 2024-10-24
Use _log_ instead of echo for logging messages

ntrrgc:
This makes it easier to disable logging messages with new --quiet options or to use syslog/journald in the future.

### DIFF
--- a/scripts/container-only/wkdev-setup-devhelp
+++ b/scripts/container-only/wkdev-setup-devhelp
@@ -43,15 +43,15 @@ run() {
     local port="$(argsparse_is_option_set 'wpe' && echo 'WPE' || echo 'GTK')"
     local docs_dir="${webkit_directory}/WebKitBuild/${port}/${build_type}/Documentation"
 
-    echo ""
+    _log_ ""
 
     if [ ! -d "${webkit_directory}/WebKit.xcworkspace" ]; then
-        echo "Passed directory does not appear to be a WebKit source directory: ${webkit_directory}"
+        _log_ "Passed directory does not appear to be a WebKit source directory: ${webkit_directory}"
         exit 1
     fi
 
     if [ ! -d "${docs_dir}" ]; then
-        echo "WebKit ${build_type} has not been built with documentation."
+        _log_ "WebKit ${build_type} has not been built with documentation."
         exit 1
     fi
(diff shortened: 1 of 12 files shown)


utilities/host-setup-tasks.sh
```diff
@@ -1,16 +1,18 @@
 #!/usr/bin/env bash
 
+[ -z "${application_ready}" ] && { echo "[FATAL] You need to source 'utilities/application.sh' before sourcing this script."; return 1; }
```
nikolaszimmermann, line 3: Is this addition intentional?

scripts/container-only/wkdev-setup-devhelp
```diff
@@ -65,18 +65,20 @@ run() {
...
-    printf "  Books:\t"
-    for dir in "${docs_dir}"/*; do
-        echo -n " $(basename "${dir}")"
-        ln --symbolic --force "${dir}" "${books_dir}/$(basename "${dir}")"
-    done
-    echo ""
-
-    echo ""
-    echo "Simply run 'devhelp' to access the documentation."
+    _log_ "$(
```
nikolaszimmermann, line 68: That seems like the only non-mechanical change? Did you check the script still works? Code wise it looks fine.

nikolaszimmermann (approved):
I assume you checked that the log method is available in all scripts that use it now?

Looks good, just a few questions.

utilities/host-setup-tasks.sh
ntrrgc: Yes, as (1) it reflects usage as far as I could find, (2) `_log_` is defined in `application.sh`, so it would fail anyway without it.

scripts/container-only/wkdev-setup-devhelp
ntrrgc: I wasn't even aware of this command existing before, but I tested it now and it seems to work.

utilities/host-setup-tasks.sh
nikolaszimmermann: Good.

ntrrgc:
> I assume you checked that the log method is available in all scripts that use it now?

I tried, hopefully didn't miss any.

PR #65, opened 2024-10-24
Add wkdev-setup-default-clang script

TingPing:
This allows easily installing and switching between different clang toolchains

### DIFF
--- a/images/wkdev_sdk/Containerfile
+++ b/images/wkdev_sdk/Containerfile
@@ -123,10 +123,12 @@ COPY /rootfs/usr/bin/podman-host /usr/bin/podman-host
 
 COPY /rootfs/etc/ccache.conf /etc/ccache.conf
 
-# Convenience symlink for clang tools, the VSCode extension doesn't find these by default.
-RUN for command in clang clang++ clangd clang-format clang-tidy lld lldb lldb-server lldb-vscode; do \
-        ln -s "/usr/bin/${command}-18" "/usr/local/bin/${command}"; \
-    done && ln -s "/usr/bin/lld-18" "/usr/local/bin/ld.lld";
+# Convenience symlinks for clang tools, the VSCode extension doesn't find these by default.
+# FIXME: Reduce code duplication with `wkdev-set-default-clang`.
+RUN bash -c 'for binary in /usr/bin/*-18; do \
+        binary_name="$(basename $binary)"; \
+        ln -s "${binary}" "/usr/local/bin/${binary_name::-3}"; \
+    done'
 
 # Fix Qt6 system packages - missing symlinks in the Ubuntu-provided packages.
 RUN export QT_VERSION=$(qmake6 -query QT_VERSION) && \
--- a/scripts/container-only/wkdev-setup-default-clang
+++ b/scripts/container-only/wkdev-setup-default-clang
@@ -1,0 +1,67 @@
+#!/usr/bin/env bash
+# Copyright 2024 Igalia S.L.
+# SPDX-License: MIT
+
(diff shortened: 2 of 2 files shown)


images/wkdev_sdk/Containerfile
```diff
@@ -127,10 +127,12 @@ COPY /rootfs/usr/bin/podman-host /usr/bin/podman-host
 
 COPY /rootfs/etc/ccache.conf /etc/ccache.conf
 
-# Convenience symlink for clang tools, the VSCode extension doesn't find these by default.
-RUN for command in clang clang++ clangd clang-format clang-tidy lld lldb lldb-server lldb-vscode; do \
-        ln -s "/usr/bin/${command}-18" "/usr/local/bin/${command}"; \
-    done && ln -s "/usr/bin/lld-18" "/usr/local/bin/ld.lld";
+# Convenience symlinks for clang tools, the VSCode extension doesn't find these by default.
+# FIXME: Reduce code duplication with `wkdev-set-default-clang`.
```
nikolaszimmermann, line 131: We should be able to invoke this script from here, no? Maybe at a later point in the script, but still..

scripts/container-only/wkdev-setup-default-clang
```diff
@@ -0,0 +1,64 @@
...
+
+run() {
+
+    argsparse_parse_options "${@}"
+    local version="${program_options["version"]}"
+
+    echo ""
+
+    # Sanity check versions Ubuntu actually has.
+    if (( $version < 14 )) || (( $version > 18)); then
```
nikolaszimmermann, line 36: I'm a fan of `${foo}` over `$foo` -- can you adapt to keep things consistent wrt to the other scripts?

nikolaszimmermann, line 36: Maybe introduce some minVersion/maxVersion variables in the pre-amble? Makes it easier to find and not forget about the other place that now use 14/18 (help text below)

scripts/container-only/wkdev-setup-default-clang
```diff
@@ -0,0 +1,64 @@
...
+        echo "$version is not a valid value (between 14-18)."
```
nikolaszimmermann, line 37: Following @ntrrgc other PR #64 we should use `_log_` instead of `echo` here.

scripts/container-only/wkdev-setup-default-clang
```diff
@@ -0,0 +1,64 @@
...
+        if ! sudo apt-get install "clang-tools-$version" "clangd-$version" "clang-format-$version" "clang-tidy-$version" "lld-$version" "lldb-$version"; then
+            echo ""
+            echo "Failed to install clang toolchain"
+            exit 1
+        fi
+    fi
+
+    local output_path
+    if [ "$EUID" -eq 0 ]; then
+        output_path='/usr/local/bin'
```
nikolaszimmermann, line 53: You could either omit the single quotes, or use double-quotes as we do for all the other strings, unless we do need to avoid variable replacments within the strings.

images/wkdev_sdk/Containerfile
nikolaszimmermann: As long as you copy and remove in the same layer, there is no increase an image size -- I agree looks odd, but helps reducing code duplication...

nikolaszimmermann: @TingPing ping?

TingPing: I was just never happy with either option.

Id prefer to merge it as is, if you don’t oppose.

nikolaszimmermann: Go ahead ☺️

PR #68, opened 2024-10-29
wkdev-enter: quiet mode and related options for scripts

ntrrgc:
This patch adds an optional --quiet flag for wkdev-enter that suppresses logging to stdout, so that scripts can be written that use wkdev-enter without pollution from wkdev logs.

The _log_ function now calls the logger command that writes to journald. This is meant to help with troubleshooting scripts that use the quiet mode.

This patch also adds the flags --no-tty and --no-interactive, which disable --tty and --interactive in podman exec respectively, and are of interest in scripts.


nikolaszimmermann (approved):
Enthusiastic r=me, great patch

PR #69, opened 2024-10-31
Bump to GStreamer 1.24.9

philn:


images/wkdev_sdk/jhbuild/webkit-sdk-deps.modules
```diff
@@ -126,15 +126,14 @@
             version="1.4.2" />
   </meson>
 
-  <meson id="gst-plugins-rs" mesonargs="-Ddoc=disabled -Dexamples=disabled -Dtests=disabled">
+  <meson id="gst-plugins-rs" mesonargs="-Ddoc=disabled -Dexamples=disabled -Dtests=disabled -Daws=disabled -Dwebrtc=disabled -Dwebrtchttp=disabled">
```
aperezdc, line 129: Why is WebRTC support disabled? We have WebKit bots building with gst-webrtc, and the bots are to be migrated from the Flatpak SDK to the container SDK... so I imagine we would need to keep this enabled?

philn: No, that webrtc plugin ships the webrtcsrc and sink elements, which we do not use in WebKit.

aperezdc: Ah, understood.  Thanks for the explanation.

aperezdc (approved):

PR #73, opened 2024-11-07
gstreamer: Include additional patches

philn:
Out of the list only
0009-rtpfunnel-Ensure-segment-events-are-forwarded-after-.patch is going to ship in 1.24(.10). The others will ship in 1.26.


dpino (approved):
Nitpick: When applying a patch to a git repository, the argument `strip` is not actually used.

The argument `strip` is used to compose the command `patch -p%d < "%s"` (https://gitlab.gnome.org/GNOME/jhbuild/-/blob/master/jhbuild/versioncontrol/tarball.py?ref_type=heads#L239), which is used to apply a patch to a tarball. When the repository is a git repository, patches are applied via `git am` or `git apply` without using `strip`.

In any case, leaving `strip` doesn't hurt because if someone in the future modifies this repo to a tarball the patches may fail to apply.

PR #76, opened 2024-11-13
wkdev-create: Add option 'no-pid'

dpino:
When set, PID namespace is not shared between host and container.


nikolaszimmermann:
I don't think we'll need this functionality, since we won't deploy containers using these scripts here.

PR #86, opened 2025-01-16
Install VSCodium and make host .application file

justinmichaud:
- Work around lldb python issues
- Install VSCodium by default
- Add helpful clangd default settings
- Make VSCode .application file by default


scripts/container-only/wkdev-setup-vscode
```diff
@@ -96,15 +133,78 @@ install_extensions() {
...
+    fi
+}
+
+install_xdg() {
+    tee /host/${HOME}/.local/share/applications/codium.desktop << HERE
+[Desktop Entry]
+Name=VSCode WKDev
+Comment=Code Editing. Redefined.
+GenericName=Text Editor
+Exec=bash -c "source /home/${USER}/.bashrc; wkdev enter -I -e -- $CODE_EXEC"
```
TingPing, line 172:
```suggestion
Exec=bash -c "source /home/${USER}/.bashrc; wkdev-enter --exec --no-interactive -- $CODE_EXEC --unity-launch %F"
```

code takes a list of files.

explict args are more readable.

TingPing: This makes a strong assumption users configure bash, which just isn't true.

This needs to be changed to be an absolute path to `wkdev-enter` on the host. I'm not sure how to get that information off-hand.

scripts/container-only/wkdev-setup-vscode
```diff
@@ -96,15 +133,78 @@ install_extensions() {
...
+Icon=vscode-wkdev
+Type=Application
+StartupNotify=false
+StartupWMClass=VSCodium
```
TingPing, line 176: Hardcoded to vscodium only.

scripts/container-only/wkdev-setup-vscode
```diff
@@ -96,15 +133,78 @@ install_extensions() {
...
+Categories=TextEditor;Development;IDE;
+MimeType=text/plain;inode/directory;application/x-codium-workspace;
+Keywords=vscodium;codium;vscode;
+
+HERE
+    echo "Installed VSCode host launcher"
+}
+
+install_icon() {
+    cp /usr/share/pixmaps/vscod*.png /host/${HOME}/.local/share/icons/vscode-wkdev.png
```
TingPing, line 186: This should go in a size appropriate subdirectory, e.g. `~/.local/share/icons/hicolor/256x256/apps/`

scripts/container-only/wkdev-setup-vscode
```diff
@@ -96,15 +133,78 @@ install_extensions() {
...
+GenericName=Text Editor
+Exec=bash -c "source /home/${USER}/.bashrc; wkdev enter -I -e -- $CODE_EXEC"
+Icon=vscode-wkdev
+Type=Application
+StartupNotify=false
+StartupWMClass=VSCodium
+Categories=TextEditor;Development;IDE;
+MimeType=text/plain;inode/directory;application/x-codium-workspace;
+Keywords=vscodium;codium;vscode;
+
```
TingPing, line 180: Just to match the normal desktop file:

```ini
[Desktop Action new-empty-window]
Name=New Empty Window
Exec=${SAME EXEC AS ABOVE MINUS UNITY} --new-window %F
Icon=vscode-wkdev
```

scripts/container-only/wkdev-setup-vscode
```diff
@@ -96,15 +133,78 @@ install_extensions() {
...
+}
+HERE
+        echo "Installed default VSCode settings to ${VSCODE_CONFIG_PATH}."
+    else
+        echo "There was already a VSCode settings.json (${VSCODE_CONFIG_PATH}), skipping."
+    fi
+}
+
+install_xdg() {
+    tee /host/${HOME}/.local/share/applications/codium.desktop << HERE
```
TingPing, line 167: Hardcoded name referring to vscodium.

scripts/container-only/wkdev-setup-vscode
```diff
@@ -96,15 +133,78 @@ install_extensions() {
...
+    ],
+    "terminal.integrated.scrollback": 1000000000,
+    "editor.renderWhitespace": "trailing",
+    "window.menuBarVisibility": "toggle",
+    "workbench.colorCustomizations": {
+    "editorWhitespace.foreground": "#FF0000",
+    "editorWhitespace.background": "#FF0000"
+},
+"terminal.integrated.allowChords": false,
+}
```
TingPing, line 158: Some of these are pretty arbitrary. I like the `clangd.arguments` but maybe lets move this to a different PR to discuss.

scripts/container-only/.wkdev-init
```diff
@@ -239,6 +253,7 @@ TASKS=(
     "try_setup_dockerenv_file"
     "try_setup_permissions_jhbuild_directory"
     "try_setup_permissions_rust_directory"
+    "try_fix_lldb_py"
```
TingPing, line 256: Please move this to its own PR.

scripts/container-only/wkdev-setup-vscode
```diff
@@ -17,8 +17,10 @@ verify_executables_exist curl
 argsparse_allow_no_argument true
 argsparse_use_option "=yes" "Assume yes for all prompts."
 argsparse_use_option "no-extensions" "Don't install extensions."
+argsparse_use_option "proprietary" "Use proprietary VSCode."
```
TingPing, line 20: I understand people have strong opinions on this but I would prefer the default being normal vscode.

I can accept vscodium as an opt-in option but the reality is it will be less tested.

TingPing (changes requested):

scripts/container-only/wkdev-setup-vscode
justinmichaud: The clangd and fg/bg colours are important for our style guidelines, but I removed the rest.

justinmichaud:
Thank you for the review, everything has been updated!

scripts/container-only/wkdev-setup-vscode
TingPing: I'm not sure I know what guidelines you refer to

TingPing: Oh you just mean to more easily visualize whitespace.

I've never needed that and don't think I'd want it by default. I'll check how it looks.

scripts/container-only/wkdev-setup-vscode
```diff
@@ -51,15 +61,50 @@ install_vscode() {
...
+        exit 1
+    fi
+
+    _log_ ""
+    _log_ "Visual Studio Code (oss) has been installed."
+}
+
 install_extension() {
 
+    sudo chown "${USER}" "${HOME}/.config"
```
TingPing, line 99: How would .config be owned by a different user?

justinmichaud: Maybe this was an idiosyncrasy of my setup, let's remove it and see if it comes up again

scripts/container-only/wkdev-setup-vscode
TingPing: I'm going to vote no on this, I think it would annoy most people.

image

justinmichaud: Oh shoot, it doesn't look like that for me. It is only supposed to display trailing whitespace

justinmichaud: 2025-01-27T13:27:24,752795531-07:00

scripts/container-only/wkdev-setup-vscode
justinmichaud: This is needed if you run this script from your firstrun script. It seems it runs as root. I guess I can change my firstrun script to run as the container user

justinmichaud: What I said makes no sense, since we use $USER here

justinmichaud: It looks like systemd makes the config path? At the very least, some files in ~/.config/systemd appear to be owned by root.

justinmichaud: Yeah, this is confusing to me, although .config is definitely owned by root. In particular, so is .config/systemd, which seems to come from nowhere. Do you have any ideas how I  can start debugging this?

TingPing: I'm not certain, but I cannot reproduce this:

```
> wkdev-create --home=$HOME/wkdev-testing --create-home
> ls -lah ~/wkdev-testing/
total 16K
drwxr-x---. 1 tingping tingping  110 Feb  5 11:52 ./
drwx------. 1 tingping tingping 2.5K Feb  5 11:52 ../
-rw-r--r--. 1 tingping tingping   18 Feb  5 11:52 .bash_logout
-rw-r--r--. 1 tingping tingping  144 Feb  5 11:52 .bash_profile
-rw-r--r--. 1 tingping tingping  522 Feb  5 11:52 .bashrc
drwxr-xr-x. 1 tingping tingping    8 Feb  5 11:52 .config/
-rw-r--r--. 1 tingping tingping  168 Feb  5 11:52 .gdbinit
drwxr-xr-x. 1 tingping tingping   34 Feb  5 11:52 .mozilla/
> ls -lah ~/wkdev-testing/.config/
total 0
drwxr-xr-x. 1 tingping tingping   8 Feb  5 11:52 ./
drwxr-x---. 1 tingping tingping 110 Feb  5 11:52 ../
drwxr-xr-x. 1 tingping tingping  22 Feb  5 11:52 fish/
```

The fact `.mozilla` exists is interesting. There must be some template being copied for a new users home.

TingPing: Could you test: https://github.com/Igalia/webkit-container-sdk/pull/98

justinmichaud: Sure!

scripts/container-only/wkdev-setup-vscode
```diff
@@ -96,15 +141,118 @@ install_extensions() {
...
+    fi
+}
+
+install_xdg() {
+    tee "/host/${HOME}/.local/share/applications/code-wkdev.desktop" << HERE
+[Desktop Entry]
+Name=VSCode WKDev
+Comment=Code Editing. Redefined.
+GenericName=Text Editor
+Exec="${WKDEV_SDK_HOST}/scripts/host-only/wkdev-enter" --exec --no-interactive -- $CODE_EXEC %F
```
justinmichaud, line 177: This doesn't quite work, we need to set WKDEV_SDK

justinmichaud:
I have simplified this PR:

- We always use the tar version of vscode or vscodium, so that this can work for armv7, arm64 and x64
- Only the settings that were not contentious are set
- I got rid of a bunch of duplicated code.

This is tested and working on arm64 and armv7

nikolaszimmermann:
@justinmichaud ok to merge as-is? Could you verify once again this is still working fine?
@TingPing can you have another look? Let's get this off the review queue :-)

PR #87, opened 2025-01-17
Add support for running containers with foreign architectures

justinmichaud:
This adds a -arch option for building and running the container image.

This was tested with qemu, building an armv7 image for intel. The goal however is to use it for armv7 on arm64.

When building for armv7, rustup was not found. This installs it in the containerfile as a workaround, although I am not sure how this ever worked. It would be nice to provide pre-built armv7 (and perhaps riscv) images.

### DIFF
--- a/images/wkdev_sdk/Containerfile
+++ b/images/wkdev_sdk/Containerfile
@@ -81,8 +81,6 @@ ENV RUSTUP_HOME="/opt/rust" \
     CARGO_HOME="/opt/rust" \
     PATH="/opt/rust/bin:${PATH}"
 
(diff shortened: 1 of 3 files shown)


justinmichaud:
Update: the Containerfile changes may not be needed? Although jhbuild doesn't seem to work on armv7 at the moment.

TingPing:
> It would be nice to provide pre-built armv7 (and perhaps riscv) images.

My plan was to do this as soon as Github had arm runners, but now we host our own runners.

So we can probably add a job for this now and see how it performs.

TingPing:
> Although jhbuild doesn't seem to work on armv7 at the moment.

Rather, libadwaita had a build issue. Did you ever get logs for that?

PR #90, opened 2025-01-22
Armv7-build-fix

justinmichaud:
Rustup is not available on armv7, so we install rust manually.

We also add some missing dependencies

### DIFF
--- a/images/wkdev_sdk/Containerfile
+++ b/images/wkdev_sdk/Containerfile
@@ -84,7 +84,8 @@ ENV RUSTUP_HOME="/opt/rust" \
     CARGO_HOME="/opt/rust" \
     PATH="/opt/rust/bin:${PATH}"
 
-RUN rustup default stable && \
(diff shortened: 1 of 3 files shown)


images/wkdev_sdk/Containerfile
```diff
@@ -81,7 +81,8 @@ ENV RUSTUP_HOME="/opt/rust" \
     CARGO_HOME="/opt/rust" \
     PATH="/opt/rust/bin:${PATH}"
 
-RUN rustup default stable && \
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y && \
```
TingPing, line 84: You can remove `rustup` from `images/wkdev_sdk/required_system_packages/04-devtools.lst`.

For arguments we prefer the full names rather than abbreviated versions so its more readable.

TingPing:
`gst-plugins-rs` fails to build in CI.

justinmichaud:
It seems it is failing on ToT? https://github.com/Igalia/webkit-container-sdk/pull/91 is also getting the same error

nikolaszimmermann (approved):

PR #91, opened 2025-01-22
Work around lldb python issues

justinmichaud:
This allows the python installation to work with lldb scripts in the WebKit repo, like dump-class-layout.

### DIFF
--- a/scripts/container-only/.wkdev-init
+++ b/scripts/container-only/.wkdev-init
@@ -204,6 +204,20 @@ try_setup_permissions_rust_directory() {
     chown --recursive "${container_user_name}:${container_group_name}" "${rust_directory}" &>/dev/null
 }
 
+# Works around python issues when running scripts like this:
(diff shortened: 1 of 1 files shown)


scripts/container-only/.wkdev-init
```diff
@@ -204,6 +204,20 @@ try_setup_permissions_rust_directory() {
...
+# Works around python issues when running scripts like this:
+# python3 Tools/Scripts/dump-class-layout -t WebKitBuild/JSCOnly/Debug/lib/libJavaScriptCore.so JavaScriptCore JSC::JSCell
+try_fix_lldb_py() {
+
+    local user_home
+
+    user_home=$(getent passwd "${container_user_name}" | cut -d: -f6)
+
+    sudo apt install lsb-release wget software-properties-common gnupg -y
+    wget -qO - https://apt.llvm.org/llvm.sh | sudo bash -s 19 -y
```
TingPing, line 216: 1. Installing core packages doesn't belong in `.wkdev-init`, it would go in the dockerfile.
2. Downloading and installing the entire llvm toolchain is a pretty massive change... Are you saying Ubuntu's clang 18 is impossible to get working?

TingPing (changes requested):

PR #94, opened 2025-01-28
Add script to install Zed inside the container

justinmichaud:
In the interest of supporting more editors, this patch adds a script to install Zed. It would be nice for all of the most common editors to be supported, but certainly the Free ones should be included.

In general, I would love to know the stance of this project on including features like this. I don't use or have a licence for Sublime or CLion, although I would vote that this repository should support using them if somebody wants to submit a patch. That is, I think we should support the set of things that anybody might use, rather than the subset that everybody uses. For example, I am not a Vim/Emacs user, but I would see value personally if someone shared a good default configuration that worked in the container with clangd.

At the same time, I understand this might cause some dramatic scope creep / maintenance overhead. 

In particular, for this PR, Zed requires workarounds in the container to get GPU access. For that reason, regardless of the result of the discussion above, I think this install script can be included without setting the precedent that every editor can be included.

This should land after the VSCodium PR. I didn't bother to duplicate that PR since I assumed any feedback on that PR would immediately apply here.

### DIFF
--- a/scripts/container-only/wkdev-setup-vscode
+++ b/scripts/container-only/wkdev-setup-vscode
@@ -8,6 +8,14 @@ else
     echo "Please set \${WKDEV_SDK} to point to the root of the wkdev-sdk checkout."
     exit 1
 fi
+if [[ -z ${WKDEV_SDK_HOST} ]]; then
+    echo "Please set \${WKDEV_SDK_HOST} to point to the the wkdev-sdk checkout (from the host perspective). This is used when creating .desktop entries."
+    exit 1
+fi
+if [[ ! -d /host/${HOME} ]]; then
+    echo "Host and container home must have the same path: /host/${HOME}"
+    exit 1
+fi
 source "${WKDEV_SDK}/utilities/prerequisites.sh"
 
 init_application "${0}" "Configures Visual Studio Code." container-only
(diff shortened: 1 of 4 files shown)


TingPing:
> In the interest of supporting more editors

I'm not sure this is a goal. The purpose of the SDK was to standardize the developer experience.

Rather what I would prefer, and started a while back but didn't finish, is to make a generic command like `wkdev-export-application` or similar to effectively do what `install_xdg()` does here in a more generic way.

If we could make that function generic at least that would meant these editor specific commands become more simple at least, so less of a burden to support more.

> In particular, for this PR, Zed requires workarounds in the container to get GPU access.

You mean installing `vulkan-tools`? The container should already have mesa, vscode needs it too, but we can add that if needed.

justinmichaud:
> If we could make that function generic at least that would mean these editor specific commands become more simple at least, so less of a burden to support more.

I think this makes sense. Although I still like the idea of being able to pick up a known-good config for a popular editor as a starting point for new contributors.

> You mean installing vulkan-tools? The container should already have mesa, vscode needs it too, but we can add that if needed.

Yeah, if there are no objections we can just do that. I figured it might be too much

TingPing:
> Although I still like the idea of being able to pick up a known-good config for a popular editor as a starting point for new contributors

I'd be ok with that if the hard part was shared (exports).

PR #95, opened 2025-01-30
Update sccache version to 0.9.1

mbrodesser-Igalia:
Didn't try this but it should work, given cargo supports version 0.9.1.
The latter is confirmed by `cargo info sccache`.

This potentially fixes incremental build bugs for WebKit, see
<https://bugs.webkit.org/show_bug.cgi?id=286676>.

### DIFF
--- a/images/wkdev_sdk/Containerfile
+++ b/images/wkdev_sdk/Containerfile
@@ -83,7 +83,7 @@ ENV RUSTUP_HOME="/opt/rust" \
 
 RUN rustup default stable && \
     rustup component remove rust-docs && \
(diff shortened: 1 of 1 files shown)


mbrodesser-Igalia:
CC @pgorszkowski-igalia

mbrodesser-Igalia:
In order to ensure this doesn't break anything, before merging I'd build the SDK, create a container and run `wkdev-setup-sccache`.

aperezdc (approved):
Seems like a reasonable change to me. I would wait a few hours and if nobody has concerns we can merge it.

cc/ @nikolaszimmermann and @TingPing

PR #96, opened 2025-02-05
gstreamer: oggdemux: fix racy decode error

calvaris:


calvaris:
I added @philn 's patch to the SDK to ensure tests at bug https://bugs.webkit.org/show_bug.cgi?id=284370 pass.

calvaris:
There's already #97 , so this is wrong and makes no sense.

PR #107, opened 2025-03-21
Support multi-arch builds with separate workers

justinmichaud:
For now, we only build the intel image. I have locally tested that this works with arm64 too.

- Always tag images, so that we can easily tar and untar them
- Download images from the build${arch} job, and create a multi-arch manifest

### DIFF
--- a/.github/workflows/wkdev-sdk.yml
+++ b/.github/workflows/wkdev-sdk.yml
@@ -10,8 +10,8 @@ defaults:
   run:
     shell: bash
 jobs:
-  build:
-    runs-on: self-hosted
+  build_amd64:
+    runs-on: [self-hosted, x64]
     steps:
       - name: Set tag name
         run: |
@@ -29,21 +29,22 @@ jobs:
       - name: Build image
         run: |
           source ./register-sdk-on-host.sh
-          wkdev-sdk-bakery --mode=build --verbose
-          wkdev-sdk-bakery --mode=export --verbose
+          wkdev-sdk-bakery --mode=build --verbose --arch amd64
+          podman image list
+          wkdev-sdk-bakery --mode=export --verbose --arch amd64
 
       - name: Archive image
         uses: actions/upload-artifact@v4
         with:
-          name: wkdev-sdk.tar
-          path: wkdev-sdk.tar
+          name: wkdev-sdk-amd64.tar
+          path: wkdev-sdk-amd64.tar
           retention-days: 7
 
       - name: Test image
         run: |
           CONTAINER="wkdev-$(date +%s)"
           source ./register-sdk-on-host.sh
-          wkdev-create --create-home --home ${HOME}/${CONTAINER}-home --verbose --attach --no-pull --name ${CONTAINER}
+          wkdev-create --create-home --home ${HOME}/${CONTAINER}-home --verbose --attach --no-pull --name ${CONTAINER} --arch amd64
           wkdev-enter -n ${CONTAINER} --exec -- git clone --depth=1 https://github.com/WebKit/WebKit.git
           wkdev-enter -n ${CONTAINER} --exec -- ./WebKit/Tools/Scripts/build-webkit --wpe --release --generate-project-only
           wkdev-enter -n ${CONTAINER} --exec -- ./WebKit/Tools/Scripts/build-webkit --gtk --release --generate-project-only
@@ -53,7 +54,7 @@ jobs:
 
   deploy:
     runs-on: self-hosted
-    needs: build
+    needs: [build_amd64]
     if: github.ref_name == 'main' || startsWith(github.ref_name, 'tag/')
     steps:
       - name: Set tag name
@@ -62,6 +63,7 @@ jobs:
             echo "WKDEV_SDK_TAG=latest" >> "${GITHUB_ENV}"
           fi
           echo "WKDEV_SDK_CONTAINER_REGISTRY_USER_NAME=$(echo ${GITHUB_REPOSITORY_OWNER} | tr '[:upper:]' '[:lower:]')" >> "${GITHUB_ENV}"
+          echo "REPO=ghcr.io/$(echo ${GITHUB_REPOSITORY_OWNER} | tr '[:upper:]' '[:lower:]')/wkdev-sdk" >> "${GITHUB_ENV}"
 
       - name: Install podman
         run: sudo apt-get update && sudo apt-get -y install podman fuse-overlayfs
@@ -69,14 +71,18 @@ jobs:
       - name: Checkout repo
         uses: actions/checkout@v4
 
-      - name: Download image
+      - name: Download images
         uses: actions/download-artifact@v4
         with:
-          name: wkdev-sdk.tar
+          pattern: wkdev-sdk-*
+          merge-multiple: true
+      - run: ls -al
 
       - name: Deploy image
         run: |
-          podman load < wkdev-sdk.tar
+          podman load < ./wkdev-sdk-amd64.tar
+          podman image list
           echo "${{ secrets.GITHUB_TOKEN }}" | podman login ghcr.io --username=${GITHUB_REPOSITORY_OWNER} --password-stdin
           source ./register-sdk-on-host.sh
-          wkdev-sdk-bakery --mode=deploy --verbose
+          wkdev-sdk-bakery --mode=deploy --verbose --multiarch="${GITHUB_RUN_ID}_amd64"
+          podman image list
--- a/scripts/host-only/wkdev-sdk-bakery
+++ b/scripts/host-only/wkdev-sdk-bakery
@@ -22,7 +22,8 @@ argsparse_use_option =env:        "Environment variable as string array, e.g. -e
 argsparse_use_option =mode:       "Operation mode: 'build', 'deploy', or 'export'" mandatory
 argsparse_use_option idle-cores:  "Number of CPU cores to leave idle, when building the image" type:uint default:2
 argsparse_use_option =tag:        "Tag to use for created image." default:$(get_default_container_tag)
-argsparse_use_option =arch:       "Container architecture."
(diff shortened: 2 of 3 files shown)


.github/workflows/wkdev-sdk.yml
```diff
@@ -10,14 +10,12 @@ defaults:
   run:
     shell: bash
 jobs:
-  build:
-    runs-on: self-hosted
+  buildintel:
```
nikolaszimmermann, line 13: build-amd64?

.github/workflows/wkdev-sdk.yml
```diff
@@ -10,14 +10,12 @@ defaults:
...
+    runs-on: [self-hosted, x64]
```
nikolaszimmermann, line 14: Why should it run on both self-hosted==amd64, and x64 (I guess that's the name of your aarch64 runner?). Both names for the runners are bad, I'd prefer something like linux-amd64, linux-aarch64, etc.

.github/workflows/wkdev-sdk.yml
```diff
@@ -10,14 +10,12 @@ defaults:
...
     steps:
       - name: Set tag name
         run: |
-          if [ "${GITHUB_BASE_REF}" = 'main' ]; then
-            echo "WKDEV_SDK_TAG=latest" >> "${GITHUB_ENV}"
-          fi
+          echo "WKDEV_SDK_TAG=intel" >> "${GITHUB_ENV}"
```
nikolaszimmermann, line 18: Same here s/intel/amd64/ IHMO.

.github/workflows/wkdev-sdk.yml
```diff
@@ -10,14 +10,12 @@ defaults:
...
-    runs-on: self-hosted
+  build_amd64:
+    runs-on: [self-hosted==amd64]
     steps:
       - name: Set tag name
         run: |
-          if [ "${GITHUB_BASE_REF}" = 'main' ]; then
-            echo "WKDEV_SDK_TAG=latest" >> "${GITHUB_ENV}"
-          fi
+          echo "WKDEV_SDK_TAG=amd64" >> "${GITHUB_ENV}"
```
TingPing, line 18: This isn't correct, the tag had a purpose for different branches.

You could append the arch to the existing tag.

.github/workflows/wkdev-sdk.yml
```diff
@@ -31,12 +29,13 @@ jobs:
           source ./register-sdk-on-host.sh
           wkdev-sdk-bakery --mode=build --verbose
           wkdev-sdk-bakery --mode=export --verbose
+          mv wkdev-sdk.tar wkdev-sdk-amd64.tar
```
TingPing, line 32: Update the export script to use the correct naming.

.github/workflows/wkdev-sdk.yml
```diff
@@ -53,30 +52,35 @@ jobs:
...
 
       - name: Deploy image
         run: |
-          podman load < wkdev-sdk.tar
+          podman load < ./wkdev-sdk-amd64.tar
+          podman image list
           echo "${{ secrets.GITHUB_TOKEN }}" | podman login ghcr.io --username=${GITHUB_REPOSITORY_OWNER} --password-stdin
-          source ./register-sdk-on-host.sh
-          wkdev-sdk-bakery --mode=deploy --verbose
+          podman manifest create $REPO:$WKDEV_SDK_TAG
```
TingPing, line 82: I assume `WKDEV_SDK_TAG` should be an arch agnostic name here. So just `latest` for example?

.github/workflows/wkdev-sdk.yml
```diff
@@ -53,30 +52,35 @@ jobs:
...
+          for IMGTAG in amd64; do
+              podman manifest add $REPO:$WKDEV_SDK_TAG containers-storage:$REPO:$IMGTAG;
+          done
+          podman manifest push --all $REPO:$WKDEV_SDK_TAG
```
TingPing, line 86: Moving logic from `wkdev-sdk-bakery` to this CI file is not great for long term maintainance.

Everything should be easily run-able locally and on CI.

So modify the scripts to do what we need, adding options as needed.

.github/workflows/wkdev-sdk.yml
nikolaszimmermann: It makes sense to keep the architecture ouf of the tag, since we plan to publish multi-arch containers anyhow.

.github/workflows/wkdev-sdk.yml
```diff
@@ -10,14 +10,12 @@ defaults:
   run:
     shell: bash
 jobs:
-  build:
-    runs-on: self-hosted
+  build_amd64:
+    runs-on: [self-hosted==amd64]
```
nikolaszimmermann, line 14: That's neat, I wasn't aware that it's trivial to select the arch this way.

.github/workflows/wkdev-sdk.yml
nikolaszimmermann: We could add a `--arch` option to `wkdev-sdk-bakery`?

.github/workflows/wkdev-sdk.yml
justinmichaud: No, I checked the docs. The original (self-hosted, x64) was correct

.github/workflows/wkdev-sdk.yml
```diff
@@ -10,14 +10,12 @@ defaults:
...
-    runs-on: self-hosted
+  build_amd64:
+    runs-on: [self-hosted, x64]
     steps:
       - name: Set tag name
         run: |
-          if [ "${GITHUB_BASE_REF}" = 'main' ]; then
-            echo "WKDEV_SDK_TAG=latest" >> "${GITHUB_ENV}"
-          fi
+          echo "WKDEV_SDK_TAG=${GITHUB_RUN_ID}" >> "${GITHUB_ENV}"
```
TingPing, line 18: You can't override the tag like this, because the tag has meaning.

We deploy tags based on the branch name: https://github.com/Igalia/webkit-container-sdk/commit/fbf418af5e8dfe0a456b7f8fcc539cc3d11b270f

.github/workflows/wkdev-sdk.yml
```diff
@@ -62,21 +60,25 @@ jobs:
...
 
       - name: Deploy image
         run: |
-          podman load < wkdev-sdk.tar
+          podman load < ./wkdev-sdk-amd64.tar
+          podman image list
           echo "${{ secrets.GITHUB_TOKEN }}" | podman login ghcr.io --username=${GITHUB_REPOSITORY_OWNER} --password-stdin
-          source ./register-sdk-on-host.sh
-          wkdev-sdk-bakery --mode=deploy --verbose
+          wkdev-sdk-bakery --mode=deploy --verbose --multiarch="${GITHUB_RUN_ID}_amd64"
```
TingPing, line 83: I feel like `--multiarch` could figure out available images on its own rather than requiring passing an arg.

images/wkdev_sdk/required_system_packages/01-base.lst
```diff
@@ -11,4 +11,4 @@ libnvidia-egl-wayland1
 wireplumber
 
 # General utilities
-apt-file aptly atop at-spi2-core emacs git htop iotop iputils-ping kmod less libportal-dev libportal-gtk4-dev libxcb-cursor-dev locate man-db nano openssh-client strace sudo unzip vim-gtk3 wget
+apt-file aptly atop at-spi2-core emacs git htop iotop iputils-ping kmod less libportal-dev libportal-gtk4-dev libxcb-cursor-dev locate man-db nano openssh-client strace sudo unzip vim-gtk3 wget pip
```
TingPing, line 14: It's `python3-pip`.

Why is this added, why isn't its own commit?

images/wkdev_sdk/Containerfile
```diff
@@ -85,10 +85,10 @@ ENV RUSTUP_HOME="/opt/rust" \
     PATH="/opt/rust/bin:${PATH}"
 
 RUN curl --proto '=https' --tlsv1.2 --show-error --silent --fail https://sh.rustup.rs | sh -s -- -y && \
-    rustup default stable && \
+    rustup default 1.84.1 && \
```
TingPing, line 88: Put this in its own commit.

TingPing (changes requested):

nikolaszimmermann:
@justinmichaud any news here?

scripts/host-only/wkdev-sdk-bakery
```diff
@@ -88,6 +93,21 @@ deploy_image() {
 
     local image_directory="$(get_image_directory_by_name "${container_image_name}")"
     pushd "${image_directory}" &>/dev/null || _abort_ "Switching to directory '${image_directory}' failed"
+
+    if argsparse_is_option_set "multiarch"; then
+        image=$(get_qualified_name "${container_image_name}")
+        target_tag=${container_tag}
```
nikolaszimmermann, line 99: I'd add quotes here for consistency

nikolaszimmermann (approved):
Looks good just a few small questions. @TingPing please check as well, since you created the tagging logic.

TingPing (approved):

.github/workflows/wkdev-sdk.yml
```diff
@@ -29,21 +29,22 @@ jobs:
       - name: Build image
         run: |
           source ./register-sdk-on-host.sh
-          wkdev-sdk-bakery --mode=build --verbose
-          wkdev-sdk-bakery --mode=export --verbose
+          wkdev-sdk-bakery --mode=build --verbose --arch amd64
+          podman image list
```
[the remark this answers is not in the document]
justinmichaud: This is for the logs, yeah

justinmichaud:
This was reverted because it failed to deploy, I will debug the issue locally

PR #112, opened 2025-04-14
docs: Advice on limiting number of cores

ptomato:
cc @clopez — as discussed


README.md
```diff
@@ -80,6 +80,9 @@ To run tests / execute MiniBrowser, try;
 ./Tools/Scripts/run-minibrowser --wpe https://browserbench.org/MotionMark1.2/
 ```
 
+For debug builds, you should expect at least 2 GB of RAM usage per core during the build.
```
clopez, line 83: Is not only for debug builds, this tip also applies for release ones.

clopez (approved):
thanks!

PR #113, opened 2025-04-14
init: Don't require having a /etc/hosts file

TingPing:


scripts/container-only/.wkdev-init
```diff
@@ -45,7 +45,12 @@ process_command_line_arguments() {
 try_copy_hosts_file() {
 
     task_step "Copying /etc/hosts file from host to container..."
-    cp /host/etc/hosts /etc/hosts || _abort_ "Cannot copy /etc/hosts file from host to container"
+    if [ -f /host/etc/hosts ]; then    
+        cp /host/etc/hosts /etc/hosts || _abort_ "Cannot copy /etc/hosts file from host to container"
+    else
+        echo "Skipping, no hosts file."
+        touch /etc/hosts
```
clopez, line 52: mmm, rather than creating an empty host file I suggest creating one that can at least resolve localhost:

```
cat > /etc/hosts << EOF
127.0.0.1	localhost
::1	localhost ip6-localhost ip6-loopback
EOF
```

PR #121, opened 2025-05-14
image: Add LLVM APT repo

TingPing:
This allows for easily installing Clang 20


justinmichaud:
Nice! I think this might fix the issues I have been having with the various lldb python scripts

PR #123, opened 2025-05-15
Support multi-arch builds with separate workers

justinmichaud:
For now, we only build the intel image. I have locally tested that this works with arm64 too.

- Always tag images, so that we can easily tar and untar them
- Download images from the build${arch} job, and create a multi-arch manifest

### DIFF
--- a/scripts/host-only/wkdev-sdk-bakery
+++ b/scripts/host-only/wkdev-sdk-bakery
@@ -22,7 +22,8 @@ argsparse_use_option =env:        "Environment variable as string array, e.g. -e
 argsparse_use_option =mode:       "Operation mode: 'build', 'deploy', or 'export'" mandatory
 argsparse_use_option idle-cores:  "Number of CPU cores to leave idle, when building the image" type:uint default:2
 argsparse_use_option =tag:        "Tag to use for created image." default:$(get_default_container_tag)
-argsparse_use_option =arch:       "Container architecture."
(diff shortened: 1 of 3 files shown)


scripts/host-only/wkdev-sdk-bakery
```diff
@@ -88,6 +93,22 @@ deploy_image() {
...
     local image_directory="$(get_image_directory_by_name "${container_image_name}")"
     pushd "${image_directory}" &>/dev/null || _abort_ "Switching to directory '${image_directory}' failed"
+
+    if argsparse_is_option_set "multiarch"; then
+        image=$(get_qualified_name "${container_image_name}")
+        target_tag="${container_tag}"
+        image_qualified="$(get_tag_for_build)"
+        echo "Building multiarch image for ${image_qualified}"
+        input_tags="$(podman image list "${image}" --format "{{.Tag}}" | grep "${target_tag}_")"
+        run_podman_silent_unless_verbose rmi --ignore "${image_qualified}" || _abort_ "Deleting existing manifest failed"
```
justinmichaud, line 103: This is the fix for the build failure in the version of this PR that got reverted

TingPing: Why does it already exist? Shouldn't the other images have a different qualified name?

justinmichaud: No, since we reuse the 'latest' tag on main

TingPing (approved):

PR #138, opened 2025-06-06
Fix multiarch input_tag iteration

justinmichaud:


scripts/host-only/wkdev-sdk-bakery
```diff
@@ -103,7 +103,7 @@ deploy_image() {
         input_tags=${input_tags//'\n'/ }
         echo "Creating manifest for ${image_qualified}"
         run_podman_silent_unless_verbose manifest create "${image_qualified}" || _abort_ "Creating manifest failed"
-        for input_tag in "${input_tags[@]}"
+        for input_tag in $input_tags
```
nikolaszimmermann, line 106: so this is a single string now? I'm confused

PR #140, opened 2025-06-17
Modify our .gdbinit default file, to handle SIGUSR1.

nikolaszimmermann:
After attaching to WPEWebProcess, gdb would halt all the time at SIGUSR1. Make sure to set the SIGUSR1 handler to 'nostop' so it doesn't halt in gdb everytime it fires (JSC uses this for GC internally).

### DIFF
--- a/images/wkdev_sdk/user_home_directory_defaults/dot-gdbinit
+++ b/images/wkdev_sdk/user_home_directory_defaults/dot-gdbinit
@@ -6,3 +6,4 @@ set print array-indexes on
 set python print-stack full
 set debuginfod enabled on
 set sysroot /
+handle SIGUSR1 nostop noprint


images/wkdev_sdk/user_home_directory_defaults/dot-gdbinit
```diff
@@ -6,3 +6,4 @@ set print array-indexes on
 set python print-stack full
 set debuginfod enabled on
 set sysroot /
+handle signal SIGUSR1 pass nostop
```
aoikonomopoulos, line 9: Hmm, personally I use `nostop` and `noprint`. AFAIU `pass` is the default.

images/wkdev_sdk/user_home_directory_defaults/dot-gdbinit
```diff
@@ -6,3 +6,4 @@ set print array-indexes on
 set python print-stack full
 set debuginfod enabled on
 set sysroot /
+handle signal SIGUSR1 nostop noprint
```
aoikonomopoulos, line 9: Oops, didn't notice it before, there's a stray `signal` in there. The line should read `handle SIGUSR1 nostop noprint` (copy-pasted from GDB to avoid any typos).

PR #141, opened 2025-06-27
Use podman manifest push instead of podman push for multiarch images

justinmichaud:
This is required for push to push all images, instead of just the one for the current architecture.


justinmichaud:
I'll merge now, since this doesn't change the build step

PR #144, opened 2025-07-02
Always set tag names to ensure the correct image is tested

justinmichaud:


scripts/host-only/wkdev-create
```diff
@@ -441,17 +441,8 @@ build_podman_create_arguments() {
...
         echo "Overriding container architecture: ${container_arch}"
         arguments+=("--arch=${container_arch}")
-
-        if ! podman image exists "$(get_sdk_qualified_name):${container_tag}"; then
-            echo "Image $(get_sdk_qualified_name):${container_tag} does not exist, trying arch-specific version."
-            container_tag="${container_tag}_${container_arch}"
-        fi
     fi
 
-    if ! podman image exists "$(get_sdk_qualified_name):${container_tag}"; then
```
nikolaszimmermann, line 451: Why are you removing this check in general?

nikolaszimmermann (commented):
Looks much better!

scripts/host-only/wkdev-create
justinmichaud: We should pull if the image doesn't exist, unless no-pull is set. In that case, podman will fail. Since we remove the image above before building, I think we should be testing the right image.

nikolaszimmermann: But the friendly message is gone?
If someone passes a wrong tag, shouldn't we warn?

That's the only question remaining here for me.

nikolaszimmermann (approved):

justinmichaud:
This was reverted because it failed on main

PR #147, opened 2025-07-31
Skip nvidia configuration when --arch is supplied.

justinmichaud:
This fixes 32-bit containers, which must use software rendering anyway.


scripts/host-only/wkdev-create
```diff
@@ -477,7 +478,9 @@ build_podman_create_arguments() {
     try_process_icons ${1}
     try_process_fonts ${1}
     try_process_dri ${1}
-    try_process_nvidia_gpu ${1}
+    if ! argsparse_is_option_set "arch"; then
```
TingPing, line 481: Could you add a comment here. In theory nvidia supports like arm, but I would imagine nobody will ever use that so its fine.

scripts/host-only/wkdev-create
```diff
@@ -331,6 +331,7 @@ try_process_dri() {
 
 try_process_nvidia_gpu() {
 
+
```
TingPing, line 334:
```suggestion
```

TingPing (approved):

PR #148, opened 2025-08-26
Switch base image to Ubuntu 25.04.

nikolaszimmermann:


nikolaszimmermann:
I discovered three distinct issues related to GCC 14 that I already fixed upstream in preparation for the bump. I'm already using this image since a month on my main development machine without a problem, just to give some additional data points.

I was under the impression that flatpak SDK was also using quite recent versions of everything - so I didn't give this much thought before starting the bump effort.

TingPing:
> I think alone the compiler versions that are more recent are a reason to update since this is meant to be a developer SDK.

We have access to basically every clang version. So its just a newer GCC toolchain (losing the older version).

It's not my intention to fight against it, just thinking about reasons for and against.

PR #153, opened 2025-09-08
devtools: Add libappstream-dev

### LINKED ISSUE #151: Adwaita dependencies not met when building
Hello,

I'm building the container image from source for an x86 build machine. I ran into a strange issue on two different machines where `libadwaita` will fail at the configuration step:

```
524.6 meson setup --prefix /jhbuild/install --libdir lib -Dexamples=false -Dgtk_doc=false -Dtests=false -Dvapi=false --buildtype=debugoptimized /jhbuild/checkout/libadwaita 524.6 *** Error during phase configure of libadwaita: ########## Error running meson setup --prefix /jhbuild/install --libdir lib -Dexamples=false -Dgtk_doc=false -Dtests=false -Dvapi=false --buildtype=debugoptimized /jhbuild/checkout/libadwaita *** [12/17]
```

I edited the Containerfile to stop before it ran this command: https://github.com/Igalia/webkit-container-sdk/blob/main/images/wkdev_sdk/Containerfile#L105C1-L112C27

Then ran the image and ran the command manually.

After poking around a bit I found this in the logs:
```
The Meson build system Version: 1.8.1 Source dir: /jhbuild/checkout/libadwaita Build dir: /jhbuild/checkout/libadwaita/_build Build type: native build Project name: libadwaita Project version: 1.6.0

[...lines omitted for brevity...]

Run-time dependency appstream found: NO (tried pkgconfig and cmake) Looking for a fallback subproject for the dependency appstream
Looking for a fallback subproject for the dependency appstream

Executing subproject appstream

[...lines omitted for brevity...]

appstream| Run-time dependency libfyaml found: NO (tried pkgconfig and cmake) subprojects/appstream/meson.build:172:12: ERROR: Dependency "libfyaml" not found, tried pkgconfig and cmake
```

`libfyaml` seems like a pretty obscure dependency, but I did see this similar package was being installed for Adwaita already: https://github.com/Igalia/webkit-container-sdk/blob/main/images/wkdev_sdk/required_system_packages/04-devtools.lst#L23

I added `libfyaml-dev` to that `devtools.lst` file which does fix the error and the container builds and runs correctly. I didn't open a pull request because it seems very strange that no one else seems to be running into this issue. 

I'm building (and running) the container with Docker CE on Ubuntu, not Podman because that's what's setup on our build machines right now. 

I'm curious if anyone has run into similar issues in the past or if anyone knows why I might be running into it now.

philn:
Required for Adwaita.

Fixes #151

### DIFF
--- a/images/wkdev_sdk/required_system_packages/04-devtools.lst
+++ b/images/wkdev_sdk/required_system_packages/04-devtools.lst
@@ -21,6 +21,7 @@ podman
 
 # For libadwaita
 libyaml-dev
+libappstream-dev
 
 # post-install for Epiphany
 desktop-file-utils


nikolaszimmermann (approved):
Thanks Phil!

PR #170, opened 2025-10-07
Prune old containers/images older than 2 weeks.

nikolaszimmermann:
This hopefully avoids the problem with disk space increasing over time.


aoikonomopoulos (approved):
Not sure how strict github's yaml schema is; if we can freely add keys, perhaps we can define this as an anchor and reference it using an alias ([random link](https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/))? If not, looks good as it is.

PR #173, opened 2025-10-20
Add script to configure remote ccache support.

nikolaszimmermann:
In the SDK, execute `wkdev-setup-remote-ccache` once and re-login. Then `build-webkit` will utilize the remote ccache support.

Check using `ccache -s` that cache hits are visible in the remote backend section.

### DIFF
--- a/scripts/container-only/wkdev-setup-remote-ccache
+++ b/scripts/container-only/wkdev-setup-remote-ccache
@@ -1,0 +1,128 @@
+#!/usr/bin/env bash
+# Copyright 2025 Igalia S.L.
+# SPDX-License: MIT
+
+if [ -f "${WKDEV_SDK}/.wkdev-sdk-root" ]; then
+    source "${WKDEV_SDK}/utilities/application.sh"
+else
+    echo "Please set \${WKDEV_SDK} to point to the root of the wkdev-sdk checkout."
+    exit 1
+fi
+source "${WKDEV_SDK}/utilities/prerequisites.sh"
+
+init_application "${0}" "Configures ccache with remote storage backend." container-only
+
+verify_executables_exist knock
+
+argsparse_allow_no_argument true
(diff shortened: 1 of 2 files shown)


scripts/container-only/wkdev-setup-remote-ccache
```diff
@@ -0,0 +1,126 @@
...
+    cat <<EOF >>"${ccache_config}"
+remote_storage = http://${server}:${port}|layout=bazel
+reshare = true
+EOF
+
+    _log_ "Writing port knocking script to ${ccache_knock_script} (executed every time the container is entered)..."
+ 
+    cat <<EOF >>"${ccache_knock_script}"
+knock -d 500 "${server}" 1111 2222 3333 4444 5555
+EOF
```
clopez: I think is better to not unconditionally send this sequence if the port is already open because each time the sequence is sent the server executes iptables rules, and that is usually not a big deal to have the same rule repeated, but if every developer and bot is going to run this each time a webkit build starts then it may end causing an issue in the end because we are talking about hundreds of requests per day.

So, what about this ?
```
if ! curl -s http://${server}:${port}/status | grep -q MaxSize; then
  knock -d 500 "${server}" 1111 2222 3333 4444 5555
fi 
```

Also it looks the script misses the shebang (`#!/bin/bash`)

scripts/container-only/wkdev-setup-remote-ccache
```diff
@@ -0,0 +1,126 @@
...
+        exit 1
+    fi
+
+    _log_ "Adding remote_storage section to ${ccache_config}..."
+    cat <<EOF >>"${ccache_config}"
+remote_storage = http://${server}:${port}|layout=bazel
+reshare = true
+EOF
+
+    _log_ "Writing port knocking script to ${ccache_knock_script} (executed every time the container is entered)..."
```
nikolaszimmermann: comment is wrong as well, will update.

clopez (approved):

PR #191, opened 2025-11-25
Install symlinks from bare programs to python3 versions

aperezdc:
Add the python-is-python3 package to the list of installed ones. This includes unversioned symlinks point to Python 3.x binaries: `python` will be a symbolic link to `python3`, `pydoc` to `pydoc3` and so on, for the sake of convenience.


nikolaszimmermann (approved):
LGTM, thanks @aperezdc

Issue #194, opened 2025-12-01
wkdev-create does not respect various env vars for paths


TingPing:
The script hardcodes the default paths, we need to look up the env vars that set them such as `DBUS_SESSION_BUS_ADDRESS` or `SSH_AUTH_SOCK`  

- [ ] keyring
- [ ] pulseaudio
- [ ] system dbus
- [ ] session dbus

(maybe more)

PR #195, opened 2025-12-01
Respect environmental overrides for various sockets

GeorgesStavracas:
Many services provide a way to override the sockets that their respective libraries connect to. We should respect that when creating containers.

This patch covers:

 * The keyring
 * Session and system D-Bus
 * PulseAudio
 * AT-SPI

For the keyring, notice that we're using the dirname of the SSH agent socket!

Closes: https://github.com/Igalia/webkit-container-sdk/issues/194

### DIFF
--- a/scripts/host-only/wkdev-create
+++ b/scripts/host-only/wkdev-create
@@ -262,6 +262,12 @@ try_process_journal() {
 try_process_keyring() {
 
     local keyring_directory="${XDG_RUNTIME_DIR}/keyring"
+
+    # Respect $SSH_AUTH_SOCK (but strip the "unix:path=" prefix)
+    if [ -v SSH_AUTH_SOCK ]; then
+        keyring_directory=$(dirname "${SSH_AUTH_SOCK##unix:path=}")
+    fi
+
     local keyring_directory_exists=$([ -d "${keyring_directory}" ] && echo 1 || echo 0)
     local podman_argument=("--mount" "type=bind,source=${keyring_directory},destination=${keyring_directory},rslave")
     podman_argument+=("--env" "SSH_AUTH_SOCK=${keyring_directory}/ssh")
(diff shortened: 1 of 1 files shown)


scripts/host-only/wkdev-create
```diff
@@ -262,6 +262,10 @@ try_process_journal() {
 try_process_keyring() {
 
     local keyring_directory="${XDG_RUNTIME_DIR}/keyring"
+
+    # Respect $SSH_AUTH_SOCK (but strip the "unix:path=" prefix)
+    [ -v SSH_AUTH_SOCK ] && keyring_directory=$(dirname "${SSH_AUTH_SOCK##unix:path=}")
+
     local keyring_directory_exists=$([ -d "${keyring_directory}" ] && echo 1 || echo 0)
     local podman_argument=("--mount" "type=bind,source=${keyring_directory},destination=${keyring_directory},rslave")
     podman_argument+=("--env" "SSH_AUTH_SOCK=${keyring_directory}/ssh")
```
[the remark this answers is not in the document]
nikolaszimmermann: No worries, I missed the "dirname" prefix on the SSH_AUTH_SOCK above and thought it's going to evaluate to /ssh/ssh...

scripts/host-only/wkdev-create
```diff
@@ -262,6 +262,10 @@ try_process_journal() {
 try_process_keyring() {
 
     local keyring_directory="${XDG_RUNTIME_DIR}/keyring"
+
+    # Respect $SSH_AUTH_SOCK (but strip the "unix:path=" prefix)
+    [ -v SSH_AUTH_SOCK ] && keyring_directory=$(dirname "${SSH_AUTH_SOCK##unix:path=}")
```
TingPing: Logic looks good to me. I'd just request not doing one-liners and instead have an explicit `if` statement.

GeorgesStavracas: Done!

PR #197, opened 2025-12-05
Add OpenXR SDK

fujii:
Add OpenXR SDK 1.1.54

Build OpenXR SDK using jhbuild.

Adding OpenXR SDK at 1.1.54 for XR_ANDROID_RAYCAST and XR_ANDROID_trackables
supports since the version shipped by Ubuntu 24.04 LTS is too old and doesn't
include them.

<https://github.com/KhronosGroup/OpenXR-SDK/releases/tag/release-1.1.54>

### DIFF
--- a/images/wkdev_sdk/jhbuild/webkit-sdk-deps.modules
+++ b/images/wkdev_sdk/jhbuild/webkit-sdk-deps.modules
@@ -15,6 +15,7 @@
       <dep package="libbacktrace"/>
       <dep package="sysprof"/>
       <dep package="libspiel"/>
+      <dep package="openxr"/>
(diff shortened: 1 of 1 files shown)


spenap:
I think going for 1.1.54 is good, as that probably includes security fixes. I think a message like

> Build OpenXR SDK using jhbuild 
>
> Adding OpenXR SDK at 1.1.54 for XR_ANDROID_RAYCAST and XR_ANDROID_trackables support since the version shipped by Ubuntu 24.04 LTS is too old and doesn't include them.

would do, thanks, @fujii !

svillar:
I'm fine with the change

PR #220, opened 2026-03-18
jhbuild: build libbacktrace as a shared library

clopez:
* Instead of building libbacktrace as a .a file that gets embedded into the webkit binaries built it as a .so that gets linked.

* This also removes the env vars `INCLUDE` and `LIB` which are non-standard and replaces that ones with `CMAKE_PREFIX_PATH` which is preferred to tell CMake where to find libraries that it can't find via the standard pkg-config mechanism (like libbacktrace). See: https://cmake.org/cmake/help/latest/variable/CMAKE_PREFIX_PATH.html


obyknovenius (approved):
Thank you for working on this.

PR #222, opened 2026-03-19
Deprecate environment variable WEBKIT_BUILD_USE_SYSTEM_LIBRARIES.

clopez:
This is not longer needed since https://commits.webkit.org/309557@main
The env var used by the WebKit tooling to identify if is running inside the SDK is now `WEBKIT_CONTAINER_SDK`.

However, to allow building older versions of WebKit with the SDK, instead of just removing the variable now, add a comment about the deprecation status with the idea of removing it on the future.

### DIFF
--- a/images/wkdev_sdk/Containerfile
+++ b/images/wkdev_sdk/Containerfile
@@ -23,13 +23,16 @@ ARG APT_AUTOREMOVE="apt-get --assume-yes autoremove"
 # Disable prompt during package configuration
 ENV DEBIAN_FRONTEND noninteractive
 
-# Used by the WebKit tooling to detect if is running inside the SDK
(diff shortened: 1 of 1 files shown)


TingPing:
You can't immediately remove this because of a change from 10 hours ago. People will be building the past few years of commits inside the sdk.

clopez:
> You can't immediately remove this because of a change from 10 hours ago. People will be building the past few years of commits inside the sdk.

mmmm, good point.

then let's keep it but add a note about it being deprecated

clopez:
I have also marked https://github.com/WebKit/WebKit/pull/60862 as suggested to be backported to 2.50 and 2.52

nikolaszimmermann (approved):

PR #223, opened 2026-03-19
jhbuild: Add wpe-platform-gtk module

TingPing:
For testing and developing wpe-platform

### DIFF
--- a/images/wkdev_sdk/jhbuild/webkit-sdk-deps.modules
+++ b/images/wkdev_sdk/jhbuild/webkit-sdk-deps.modules
@@ -241,6 +241,13 @@
     </dependencies>
   </meson>
 
+  <!-- Optional WPE Platform implementaiton. -->
+  <meson id="wpe-platform-gtk">
+    <branch repo="github.com"
+            module="Igalia/wpe-platform-gtk.git"/>
+    <!-- This depends on gtk4, but the older distro provided version is fine. -->
+  </meson>
+
   <!-- Everything in this section is just for sysprof. -->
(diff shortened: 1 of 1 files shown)


images/wkdev_sdk/jhbuild/webkit-sdk-deps.modules
```diff
@@ -241,6 +241,13 @@
     </dependencies>
   </meson>
 
+  <!-- Optional WPE Platform implementaiton. -->
+  <meson id="wpe-platform-gtk">
+    <branch repo="github.com"
+            module="Igalia/wpe-platform-gtk.git"/>
+    <!-- This depends on gtk4, but the older distro provided version is fine. -->
```
clopez, line 248:
```
<dependencies>
 <dep package="gtk4"/>
<dependencies>
```
Why you don't add that this instead of the comment?

We are building gtk4 inside the jhbuild, we don't use the distro provided one, so adding the dep entry on jhbuild is needed to ensure proper order of building

clopez (changes requested):
Another question:
How this module is expected to be used from jhbuild?
Do you need to install first wpe on a path and then build this module?

TingPing:
>  Do you need to install first wpe on a path and then build this module?

Yeah that's what I've been doing

nikolaszimmermann:
@TingPing / @clopez any concerns left to discuss?

nikolaszimmermann:
Ping @TingPing

TingPing:
I'm just going to close this, wig uses a submodule for example.

PR #224, opened 2026-04-08
Add libvulkan-volk-dev package

aperezdc:
Volk will be used for Vulkan support in WebKit, to allow dynamically loading the Vulkan libraries at runtime and providing extension wrangling, similar to how Epoxy is used for OpenGL.


spenap (approved):
LGTM

PR #229, opened 2026-04-15
wkdev-enter: use --detach-keys=

ntrrgc:
By default podman exec binds Ctrl+p,Ctrl+q to detach. This causes Ctrl+p to not work normally (it is only send to the process once a second key is received, when podman is sure it wasn't part of a Ctrl+p,Ctrl+q sequence).

This can be disabled with `--detach-keys=""`. This patch does that.

Unfortunately, this does not work when podman --remote is used due to a bug in podman (https://github.com/containers/podman/issues/28193) so this patch will only work once that also gets fixed.

### DIFF
--- a/scripts/host-only/wkdev-enter
+++ b/scripts/host-only/wkdev-enter
@@ -75,11 +75,6 @@ build_podman_arguments() {
 
     local -n generic_arguments=${1}
 
-    # On the host, the podman user socket can be used as well:
-    # '--remote' routes podman communication through the socket,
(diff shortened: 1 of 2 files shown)


ntrrgc:
After discussing it with @nikolaszimmermann, I think it makes sense to remove the usage of `podman --remote`, since it is no longer relevant, and that way we don't have to wait for it to be fixed to have working keys.

nikolaszimmermann (approved):

PR #230, opened 2026-04-16
.gdbinit: Load pretty printers

ntrrgc:
GLib and GStreamer install pretty printers for gdb. However, since we are building both with jhbuild, they were not being found.

This patch adds the necessary code to the .gdbinit init example to load those pretty printers.

WebKit also contains (non-auto-loadable) gdb pretty printers. Since now WebKit lives in a predictable path, we can load them as well, and this is also added to the example.


nikolaszimmermann (approved):
Nice!

PR #231, opened 2026-04-17
Fix wkdev-update losing shared-dir

TingPing:
It gets the old home but not its shared-dir, so now we persist that in the metadata.

Added arguments just to help with testing, so you can do `wkdev-update --force wkdev`


nikolaszimmermann (approved):
LGTM!

PR #234, opened 2026-04-21
Add vulkan-validationlayers package

aperezdc:
These are useful during development and debugging, therefore nice to have in the SDK.


spenap (approved):
LGTM

PR #236, opened 2026-04-21
Add wkdev-find-buildbot-worker-base helper script

nikolaszimmermann:
Given a buildbot-worker version, resolves which wkdev-sdk base image digest it was built on top of (even if no longer tagged) and the source commit / GitHub Actions run that produced that base image.


nikolaszimmermann:
It's obsolete due to the work on the new versioning scheme and soon the pinning of the SDK in the WebKit tree.

PR #237, opened 2026-04-27
Introduce a proper version scheme for the wkdev-sdk image

nikolaszimmermann:
The Containerfile's new WKDEV_SDK_VERSION ARG is now the single source of truth for the SDK version. It is propagated to the OCI image label, to /etc/wkdev-sdk-version inside the container, to the welcome message, and to the default image tag used by wkdev-create / wkdev-sdk-bakery. There is no longer a :latest tag, a WKDEV_SDK_TAG environment variable, a --tag CLI flag, or a tag/* branch mechanism: to move a checkout to a different SDK version, change WKDEV_SDK_VERSION (typically by pulling the latest main).

wkdev-update is now version-aware. It prints the version pinned in the local checkout and the latest version published in the registry, then walks each wkdev-sdk-using container and flags as OUTDATED anything running an older version (or carrying no version label, which catches legacy :latest images so they can be migrated automatically). Outdated containers are recreated through wkdev-create, which picks up the new version from the Containerfile.

The CI workflow no longer publishes on every push to main; it only pushes the multi-arch image when a wkdev-sdk-X.Y.Z git tag is created (and sanity-checks that the tag matches the Containerfile version). A new release-wkdev-sdk workflow does the version bump, commit and tag in one click, then waits for the build to finish.

wkdev-sdk-bakery loses its deploy / multiarch modes since publishing now happens exclusively through the release workflow; the welcome message gained a version line, and the image's title label now mentions the Ubuntu 24.04 base so consumers can tell variants apart.


nikolaszimmermann:
This is the webkit-container-sdk-bots counterpart: https://github.com/Igalia/webkit-container-sdk-bots/pull/44

nikolaszimmermann:
Updated after todays discussions regarding the naming scheme: no more semantic versioning, but tracking WPE/WebKitGTK release cycle.

obyknovenius (approved):
I like it! Thank you for the effort.

.github/workflows/release-wkdev-sdk.yaml
```diff
@@ -0,0 +1,156 @@
...
+
+        MAJOR=$(echo "${CURRENT_VERSION}" | cut -d. -f1)
+        MINOR=$(echo "${CURRENT_VERSION}" | cut -d. -f2)
+        PATCH=$(echo "${CURRENT_VERSION}" | cut -d. -f3)
+        NEW_PATCH=$((PATCH + 1))
+        # Auto-bump the patch component only; major / minor track the upstream
+        # WebKitGTK / WPE WebKit release branches and are bumped by hand in the
+        # Containerfile when moving to a new branch.
+        NEW_VERSION="${MAJOR}.${MINOR}.${NEW_PATCH}"
+        echo "new_version=${NEW_VERSION}" >> $GITHUB_OUTPUT
```
clopez, line 44: In order to make easier to bisect regressions or to investigate issues it would be helpful if the version string includes the short sha1 git has of commit of this repository (webkit-container-sdk) that was used to build the image.

Suggested change:
```
GITSHASH="$(git log -1 --pretty=%h)"
NEW_VERSION="${MAJOR}.${MINOR}.${NEW_PATCH}-git${GITSHASH}"
```

So the version would then look like `2.53.1-git916f9ef`, making it immediately obvious that the `2.53.1` build was produced from `webkit-container-sdk` git commit `916f9ef`.

clopez: Not sure if this requires adapting some part on the code that compares versions to cut the string before the `-`.  Maybe wrap it with a function that returns it like `numeric_version() { echo ${1} | cut -d\- -f1 ; }` and then call `$(numeric_version ${VERSION})` when needed.

I think that for `is_version_greater_or_equal()` and friends is not needed, `sort -V` should sort the versions equally with that `-git${SHASH}` suffix than without it (AFAIK)

TingPing:
If I understand correctly the patch version is specific to the image, I think since it overlaps with webkit release versions there should be more clear separation from webkit `2.50.3` such as `2.50-v3` for example. (Also like Carlos comment about git hashes)

Otherwise, like the idea, looks good :+1:

.github/workflows/release-wkdev-sdk.yaml
nikolaszimmermann: It does, thanks for the suggestions.

nikolaszimmermann:
For the record:

So to create a new release for a new release cycle, we do:

1. Set `ARG WKDEV_SDK_VERSION=2.54-v1` (No `-<sha>` suffix yet -> no release has been cut for v1.) in Containerfile, git commit and git push
2. First release workflow run sees `2.54-v1` (no sha suffix) -> does not bump -> just appends sha: `2.54-v1` -> `2.54-v1-<sha>` Tag gets pushed: `wkdev-sdk-2.54-v1-<sha>`. Containerfile now holds `2.54-v1-<sha>.`
3. Second release run sees `2.54-v1-<sha>` (has sha -> "v1 was already used") -> bumps `<count>` and uses a fresh sha: `2.53-v1-<sha>` ->  `2.54-v2-<new-sha>`

So the Containerfile version will only ever iterate between those states:

1) `<M>.<m>-v<N>` - "reserved-but-unreleased" (_only_ happens after a hand-edit when preparing the first SDK release for a new WPE/WebKitGtk development cycle)
2) `<M>.<m>-v<N>-<sha>` - released (the normal state between releases)

nikolaszimmermann:
Did many further cleanups, with a few agentic simplify sessions.

PR #243, opened 2026-05-15
CI: Clean up podman storage after build/deploy jobs

nikolaszimmermann:
The wkdev-sdk image-build workflow runs on self-hosted runners on every PR but pushes to GHCR only on tags, so PR builds left their image layers behind on the runners (the existing "Cleanup test artifacts" step removed test containers and home dirs but not the built image). The deploy job likewise pulled both arch images for manifest assembly and never removed them. Both jobs now remove what they produced, run podman system prune to sweep stopped containers, dangling layers, unused networks and dangling build cache, and print labelled disk-usage reporting at the end.

The shared cleanup logic lives in a local composite action at .github/actions/podman-cleanup, so each job invokes it with only the per-job manifest/image inputs. While I'm at it, cleanup formatting, style etc to be similar to all the other workflows in webkit-container-sdk-bots.

Also adds a concurrency group that cancels superseded PR runs (push tags are not cancelled) so a rapid succession of pushes to the same PR doesn't stack multiple in-flight builds on the same runner.

This should improve the disk space situation in the future.

### DIFF
--- a/.github/workflows/build-and-publish-wkdev-sdk-image.yaml
+++ b/.github/workflows/build-and-publish-wkdev-sdk-image.yaml
@@ -7,6 +7,10 @@ on:
   pull_request:
     paths-ignore: [ docs/**, README.md ]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
   build:
     strategy:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-publish-wkdev-sdk-image.yaml
```diff
@@ -23,60 +27,62 @@ jobs:
...
+      run: echo "${{ secrets.GITHUB_TOKEN }}" | podman login ghcr.io -u ${{ github.repository_owner }} --password-stdin
+
+    - name: Extract version
+      run: |
+        VERSION=$(./scripts/helpers/print-sdk-version)
+        echo "VERSION=${VERSION}" >> $GITHUB_ENV
 
-      - name: Install podman
-        run: sudo apt-get update && sudo apt-get -y install podman fuse-overlayfs
-
```
clopez, line 48: is this gone on purpose? I guess we can do that so we avoid colliding with apt-get autoupdates.
That should be done in the base when a new runnerr is deployed for this by default, i think we don't need to keep trying to install it

.github/workflows/release-wkdev-sdk.yaml
```diff
@@ -11,6 +11,9 @@ env:
   CONTAINERFILE_PATH: images/wkdev_sdk/Containerfile
 
 jobs:
+  # ============================================================
+  # Phase 1: Image Version Bump and Tag
+  # ============================================================
```
clopez: 1-line comment is enough

.github/workflows/release-wkdev-sdk.yaml
```diff
@@ -70,6 +73,9 @@ jobs:
         git push origin "${TAG_NAME}"
         echo "Created tag: ${TAG_NAME}"
 
+  # ============================================================
+  # Phase 2: Wait for Image Build Workflow
+  # ============================================================
```
clopez: ditto

.github/workflows/release-wkdev-sdk.yaml
```diff
@@ -121,6 +127,9 @@ jobs:
 
         echo "Image build workflow completed successfully!"
 
+  # ============================================================
+  # Summary Job
+  # ============================================================
```
clopez, line 128: ditto

.github/workflows/build-and-publish-wkdev-sdk-image.yaml
```diff
@@ -87,95 +93,100 @@ jobs:
...
+        done
+
+    - name: Push self-contained multi-arch manifest with all layers
+      run: |
+        podman manifest push --all ${{ env.IMAGE_WITH_TAG }} docker://${{ env.IMAGE_WITH_TAG }}
+
+    - name: Remove local artifacts produced by this deploy
+      if: always()
+      run: |
+        podman manifest rm ${{ env.IMAGE_WITH_TAG }} 2>/dev/null || true
```
clopez, line 184: This runs even if a previous step failed `if: always()`. I think it should not do that: only clean on successful build, otherwise if something failed and a fix is pushed it has to start without the previous caches.
Also `IMAGE_WITH_TAG` can be undefined if the run failed before the step that sets it.

.github/workflows/build-and-publish-wkdev-sdk-image.yaml
```diff
@@ -23,60 +27,62 @@ jobs:
...
-          # Clean up any leftover containers and home directories from previous failed runs
-          podman ps -a --filter "name=^wkdev-" --format "{{.Names}}" | xargs -r podman rm --force 2>/dev/null || true
-          find ${HOME} -maxdepth 1 -type d -name "wkdev-*-home" -exec rm -rf {} + 2>/dev/null || true
+        fi
+
+    - name: Extract GitHub Container Registry repository name
+      run: echo "GHCR_REPO=${GITHUB_REPOSITORY_OWNER,,}" >> $GITHUB_ENV
+
+    - name: Build qualified image/tag name
+      run: echo "IMAGE_WITH_TAG=ghcr.io/${{ env.GHCR_REPO }}/wkdev-sdk:${{ env.VERSION }}" >> $GITHUB_ENV
```
clopez: This two steps are single-line echo >> $GITHUB_ENV calls. More step adds noise to the logs. The previous one-step version was better IMHO

.github/workflows/build-and-publish-wkdev-sdk-image.yaml
```diff
@@ -23,60 +27,62 @@ jobs:
...
+
+    - name: Build image
+      run: |
+        podman build --jobs $(nproc) --build-context scripts=./scripts \
+          --build-arg WKDEV_SDK_VERSION=${{ env.VERSION }} \
+          -t ${{ env.IMAGE_WITH_TAG }}_${{ matrix.arch }} --arch=${{ matrix.arch }} images/wkdev_sdk
+        podman image list
```
clopez, line 56: why this rename? env.VERSION instead of WKDEV_SDK_VERSION ?

.github/workflows/build-and-publish-wkdev-sdk-image.yaml
```diff
@@ -23,60 +27,62 @@ jobs:
...
-
-          if [[ "${GITHUB_REF}" == refs/tags/wkdev-sdk-* ]]; then
-            TAG_VERSION="${GITHUB_REF#refs/tags/wkdev-sdk-}"
-            if [ "${TAG_VERSION}" != "${VERSION}" ]; then
-              echo "Error: git tag wkdev-sdk-${TAG_VERSION} does not match Containerfile WKDEV_SDK_VERSION=${VERSION}"
-              exit 1
-            fi
-          fi
+    - name: Checkout repository
+      uses: actions/checkout@v4
```
clopez, line 31: This white-space changes make the patch harder to read. I prefer to avoid those unless necessary.

clopez (approved):

.github/workflows/build-and-publish-wkdev-sdk-image.yaml
```diff
@@ -87,95 +93,100 @@ jobs:
...
+      run: echo "${{ secrets.GITHUB_TOKEN }}" | skopeo login ghcr.io -u ${{ github.repository_owner }} --password-stdin
+
+    - name: Extract version
+      run: echo "VERSION=${GITHUB_REF#refs/tags/wkdev-sdk-}" >> $GITHUB_ENV
+
+    - name: Extract GitHub Container Registry repository name
+      run: echo "GHCR_REPO=${GITHUB_REPOSITORY_OWNER,,}" >> $GITHUB_ENV
+
+    - name: Build qualified image/tag name
+      run: echo "IMAGE_WITH_TAG=ghcr.io/${{ env.GHCR_REPO }}/wkdev-sdk:${{ env.VERSION }}" >> $GITHUB_ENV
```
clopez: this all 3 steps could be also collapsed into one

.github/workflows/release-wkdev-sdk.yaml
nikolaszimmermann: I wanted to minimize the diff to the other workflow, I need to change it in wkdev-sdk-bots then.

.github/workflows/build-and-publish-wkdev-sdk-image.yaml
nikolaszimmermann: Ok, good point!

.github/workflows/build-and-publish-wkdev-sdk-image.yaml
nikolaszimmermann: Ok, will adapt the other workflow instead.

.github/workflows/build-and-publish-wkdev-sdk-image.yaml
nikolaszimmermann: It's about minimizing the diff to the build-and-publish-buildbot-worker.yaml workflow, which uses the preferred style, this one does not. I hope it's ok this time, next time I'll do the cleanups seperated.

.github/workflows/build-and-publish-wkdev-sdk-image.yaml
nikolaszimmermann: Ok, will adapt the other workflow and this one.

.github/workflows/build-and-publish-wkdev-sdk-image.yaml
nikolaszimmermann: Only for consistency with the other workflow.

.github/workflows/build-and-publish-wkdev-sdk-image.yaml
nikolaszimmermann: Yes, it's a relict from the days where I had no access to the runners :-)

PR #247, opened 2026-05-20
Use a explicit WebKit commit for running the install-dependencies scripts.

clopez:
This allows two things:

 - Any time there is an update there to be included into the image, then trigerring a new build is as easy as updating the WebKit commit sha1 hash.
 - Be explicit with the version of the install-dependencies script that was used to build the image. This helps to make the image reproducible and helps to make more easy to bisect


nikolaszimmermann (approved):
Great idea!

PR #248, opened 2026-05-21
CI: Publish per-PR wkdev-sdk images to ghcr.io

nikolaszimmermann:
For PR branches the build/deploy pipeline now publishes a multi-arch image to ghcr.io/<owner>/wkdev-sdk tagged pr-<N>-<sha> -- coexisting with release tags (e.g. 2.53-v5-<sha>) in the same place, distinguished solely by the pr- prefix. The tag is immutable per push (no overwrites, no moving aliases), so every push gets its own pullable handle.

To make pr-tagged images launchable with the standard tooling, the `wkdev-create --version <abc>` validation now also accepts the pr-<N>-<sha> form. The regex for matching that is intentionally disjoint from WKDEV_SDK_VERSION_RE so PR images never leak into release-version listings or comparisons. The image's baked-in WKDEV_SDK_VERSION is still the canonical <MM>-v1-<sha> form, so in-container tooling continues to treat the image as a real SDK version.

On PR close/merge, .github/workflows/cleanup-wkdev-sdk-pr-tags.yaml lists all pr-<N>-* versions of the wkdev-sdk package via the GHCR API and deletes them sequentially, using the user/org type from the event payload to pick the right packages API path.

A sticky PR comment advertises the latest image and pull command. Fork PRs are excluded since their GITHUB_TOKEN lacks packages:write.

### DIFF
--- a/.github/workflows/build-and-publish-wkdev-sdk-image.yaml
+++ b/.github/workflows/build-and-publish-wkdev-sdk-image.yaml
@@ -11,6 +11,11 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
+env:
+  # Release tags (e.g. 2.53-v5-3bdf4b8) and per-PR tags (pr-<N>-<sha>) coexist
+  # in this one image; the 'pr-' prefix is the distinguisher.
+  IMAGE_NAME: wkdev-sdk
+
 jobs:
   build:
     strategy:
@@ -25,6 +30,8 @@ jobs:
     permissions:
       contents: read
       packages: write
+    outputs:
+      registry_tag: ${{ steps.resolve.outputs.registry_tag }}
 
     steps:
     - name: Checkout repository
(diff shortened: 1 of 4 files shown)


.github/workflows/build-and-publish-wkdev-sdk-image.yaml
```diff
@@ -79,28 +110,31 @@ jobs:
...
       packages: write
+      pull-requests: write
+    env:
+      REGISTRY_TAG: ${{ needs.build.outputs.registry_tag }}
 
     steps:
-    - name: Checkout repository
-      uses: actions/checkout@v4
-
     - name: Log in to GitHub Container Registry
```
clopez, line 131: why this is removed?
does the last step "Comment on PR with deployed image" from `marocchino/sticky-pull-request-comment@v2` works in a job without checkout?

clopez (approved):

.github/workflows/build-and-publish-wkdev-sdk-image.yaml
nikolaszimmermann: Yes it does, we don 't need this anymore.

PR #252, opened 2026-05-22
wkdev-sdk: Reorder Containerfile for cache reuse across releases

nikolaszimmermann:
Move every ${WKDEV_SDK_VERSION} consumer (OCI labels, /etc/wkdev-sdk-version bake) and the runtime-only ENVs to the tail, so version bumps only re-hash a few small layers instead of the multi-GB apt/rust/jhbuild stack.

The ARG WKDEV_SDK_VERSION line stays put so the release workflow keeps parsing it. Also factor apt cache cleanup into a new APT_CLEAN ARG and apply it to both apt-heavy RUNs.

### DIFF
--- a/images/wkdev_sdk/Containerfile
+++ b/images/wkdev_sdk/Containerfile
@@ -10,16 +10,6 @@ FROM public.ecr.aws/ubuntu/ubuntu:24.04
 # branch and is bumped by hand here when moving to a new branch.
 ARG WKDEV_SDK_VERSION=2.53-v5-3bdf4b8
 
-LABEL maintainer="webkit-gtk@lists.webkit.org"
-LABEL org.opencontainers.image.version=${WKDEV_SDK_VERSION}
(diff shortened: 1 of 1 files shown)


images/wkdev_sdk/Containerfile
```diff
@@ -10,16 +10,6 @@ FROM public.ecr.aws/ubuntu/ubuntu:24.04
 # branch and is bumped by hand here when moving to a new branch.
 ARG WKDEV_SDK_VERSION=2.53-v5-3bdf4b8
```
clopez, line 11: This should be moved to the bottom as well, shoudln't it?

clopez (approved):

images/wkdev_sdk/Containerfile
nikolaszimmermann: No, it matters where it is used (the RUN echo ${WKDEV_SDK_VERSION} > /etc/wkdev-sdk-version step, has to be at the bottom). It's nicer to have this at the top of the file, so I left it.

PR #256, opened 2026-06-16
Only ship a single compiler

TingPing:
We use swift and its bundled clang.

I also built extra clang tools from the same swift release.

(This is WIP and not well tested)


images/wkdev_sdk/jhbuild/webkit-sdk-deps.modules
```diff
@@ -207,6 +207,22 @@
...
+        id="clang-tools"
+        makeargs="distribution"
+        makeinstallargs="install-distribution-stripped"
+        cmakeargs="-DLLVM_ENABLE_PROJECTS='clang;clang-tools-extra' -DLLVM_DISTRIBUTION_COMPONENTS='clang-format;clang-tidy;clang-apply-replacements;clang-resource-headers;clang-headers;clang-scan-deps'"
+    >
+    <branch
+            repo="github-tarball"
+            checkoutdir="llvm-project"
+            module="swiftlang/llvm-project/archive/refs/tags/swift-${version}-RELEASE.tar.gz"
+            version="6.1.1"
```
nikolaszimmermann: We already had 6.2.3 - is that a different versioning scheme here?

images/wkdev_sdk/required_system_packages/02-gcc.lst
```diff
@@ -1,2 +0,0 @@
-# GNU GCC Toolchain (current version: 14)
-gcc-14 g++-14 gdb libc6-dev
```
nikolaszimmermann, line 2: Shouldn't we keep it for now? Also not sure if we don't have a bot depending on it right now? @clopez ?

images/wkdev_sdk/jhbuild/webkit-sdk-deps.modules
TingPing: Ah they only made a github release for 6.1.1 for some reason. But they have tags I switched to.

PR #257, opened 2026-07-06
Remove WebRTC-related dependencies

philn:
No longer needed after we switch to libwebrtc.

### DIFF
--- a/images/wkdev_sdk/Containerfile
+++ b/images/wkdev_sdk/Containerfile
@@ -142,14 +142,6 @@ RUN rustup default 1.94.0 && \
     rustup component remove rust-docs && \
     cargo install --root /usr/local --version 0.10.20 --locked cargo-c
 
-RUN git clone https://github.com/ystreet/librice.git && \
-    cd librice && \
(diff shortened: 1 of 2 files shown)


TingPing:
I believe cargo-c was just for librice too?

philn:
I think it's needed for gst-plugins-rs as well.

clopez (approved):

clopez:
It seems we have a egg-chicken problem here: the runner fails the step "test image" because WebKit is not longer anymore to complete the cmake-configure step (which is expected since you are removing libraries needed to configure current webkit).

Perhaps the easiest way to solve this is to:
 1. First land the changes on webkit to build by default with libwebrtc
 2. Then come back to this PR and retrigger the CI (this time should work since webkit now builds with libwebrtc)
 3. Land this
 4. Finally update the container version used at $WebKit/.wkdev-sdk-version

@nikolaszimmermann : any other idea?

philn:
Yes that sounds like a plan. Sorry for previous (now deleted) misleading comment.

PR #258, opened 2026-07-12
Build our own swift toolchain

TingPing:
This is a very big project to compile, so hopefully it works out.

First we build Clang/LLVM/Swift with the host tools. Then remove the host tools.

The built swift is kinda incomplete. So we have to build a few things on top like foundation libraries, llbuild, swift-driver.

I added swift-format because the WebKit style scripts depend on it and it's fairly small.

I included a few unrelated changes of just removing things we don't need to build because i was tired of slow builds :)


TingPing:
> `-DSWIFT_ENABLE_SWIFT_IN_SWIFT=OFF`

AFAICT a swift compiler is definitely a requirement but it's optional for some components. Some of the options seemed to be unhappy during configuration but didn't give any helpful error so that was disabled.

~~So to be investigated is we should build as much as we can with Swift.~~

EDIT: Fixed

clopez:
Does this means that we will build our own Clang compiler?

If that is the case then it would be awesome if we can apply LTO+PGO to the compilation so we get a very fast compiler. And if the PGO traces are obtained from a WebKit compilation that would be even more amazing.

BTW: I see that you are removing the Ubuntu clang packages from the image... I wonder how is the Clang compilation performed? It uses GCC to build it? if that is the case maybe we should use the Ubuntu Clang instead to build it? (Clang is better at LTO than GCC)

TingPing:
> BTW: I see that you are removing the Ubuntu clang packages from the image... I wonder how is the Clang compilation performed? It uses GCC to build it? if that is the case maybe we should use the Ubuntu Clang instead to build it? (Clang is better at LTO than GCC)

I install the ubuntu `swiftlang/clang` temporarily, build the whole toolchain with them, and then remove the ubuntu ones.

> If that is the case then it would be awesome if we can apply LTO+PGO to the compilation so we get a very fast compiler. And if the PGO traces are obtained from a WebKit compilation that would be even more amazing.

I've enabled LTO on everything. I think PGO is a cool idea, should be very doable, but I probably won't look into it.

TingPing:
So I've got as far as I believe I can. AFAICT the Clang toolchain is perfect, lldb, lld, sanitizers, etc.

Swift works, but there are some snags with building WebKit (I did not try any other swift toolchain to know if this is unique)


Firstly:

```
› mit-module-doc-path Source/WebKit/WebKit.swiftdoc -emit-module-source-info-path Source/WebKit/WebKit.swiftsourceinfo -emit-objc-header-path /mnt/work/webkit/WebKit/WebKitBuild/GTK/Debug/Source/WebKit/WebKit-Swift-Generated.h.tmp -module-name WebKit -o Source/WebKit/
  WebKit.swiftmodule
  1.    Swift version 6.3.3 (swift-6.3.3-RELEASE)
  2.    Compiling with the current language version
  3.    While evaluating request ASTLoweringRequest(Lowering AST to SIL for module WebKit)
  4.    While deserializing SIL function
  "$sSo3stdO00158optionalWTFObjectIdentifierGenericWebCoreFrameIdentifierTypeWTFObjectIdentifierMainThreadAccessTraitsCUnsignedLongCUnsignedLong_JEBIIctbIJhubGjEayEEpBaFeIppaaV7pointeeSo3WTFO00143ObjectIdentifierGenericWebCoreFrameIdentifierTypeWTFObjectIdentifierMainThreadAccessTr
  aitsCUnsignedLongCUnsignedLong_ehGIDctbCAhubAkyaDAobczopaVvg"
  5.    While deserializing SIL function
  "$sSo3stdO00158optionalWTFObjectIdentifierGenericWebCoreFrameIdentifierTypeWTFObjectIdentifierMainThreadAccessTraitsCUnsignedLongCUnsignedLong_JEBIIctbIJhubGjEayEEpBaFeIppaaV14__operatorStarSPySo3WTFO00143ObjectIdentifierGenericWebCoreFrameIdentifierTypeWTFObjectIdentifierMainThr
  eadAccessTraitsCUnsignedLongCUnsignedLong_ehGIDctbCAhubAkyaDAobczopaVGyFTo"
  6.    Abort: function fatal at ModuleFileSharedCore.cpp:717
  |     *** DESERIALIZATION FAILURE ***
  |     *** If any module named here was modified in the SDK, please delete the ***
  |     *** new swiftmodule files from the SDK and keep only swiftinterfaces.   ***
  |     module 'WebKit', builder version '6.3.3(6.3.3)/Swift version 6.3.3 (swift-6.3.3-RELEASE)', built from source against SDK , non-resilient, loaded from 'Source/WebKit/CMakeFiles/WebKit.dir/UIProcess/WebBackForwardList.swift.swiftmodule'
  |     result is ambiguous (_)
  |     Cross-reference to module '__ObjC'
  |     ... std
  |     ... optional
  |     ... (in c:@N@std@S@optional>#$@N@WTF@S@ObjectIdentifierGeneric>#$@N@WebCore@S@FrameIdentifierType#$@N@WTF@S@ObjectIdentifierMainThreadAccessTraits>#l#l)
  |     ... __operatorStar
  |     ... with type (std.optional<WTF.ObjectIdentifierGeneric<WebCore.FrameIdentifierType, WTF.ObjectIdentifierMainThreadAccessTraits<CUnsignedLong>, CUnsignedLong>>) -> () -> UnsafePointer<WTF.ObjectIdentifierGeneric<WebCore.FrameIdentifierType,
  WTF.ObjectIdentifierMainThreadAccessTraits<CUnsignedLong>, CUnsignedLong>>
   #0 0x00007f6fad721f78 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (/jhbuild/install/lib/libLLVM.so.21.1+0x25e8f78)
   #1 0x00007f6fad71e9f5 llvm::sys::RunSignalHandlers() (/jhbuild/install/lib/libLLVM.so.21.1+0x25e59f5)
   #2 0x00007f6fad7226b1 (/jhbuild/install/lib/libLLVM.so.21.1+0x25e96b1)
   #3 0x00007f6faab69cb0 (/usr/lib/x86_64-linux-gnu/libc.so.6+0x45cb0)
   #4 0x00007f6faabca48c __pthread_kill_implementation ./nptl/pthread_kill.c:44:76
   #5 0x00007f6faabca48c __pthread_kill_internal ./nptl/pthread_kill.c:89:10
   #6 0x00007f6faabca48c pthread_kill ./nptl/pthread_kill.c:100:10
   #7 0x00007f6faab69b7e raise ./signal/../sysdeps/posix/raise.c:27:6
   #8 0x00007f6faab4c8ec internal_signal_block_all ./stdlib/../sysdeps/unix/sysv/linux/internal-signals.h:81:3
   #9 0x00007f6faab4c8ec __abort_lock_wrlock ./stdlib/abort.c:62:3
  #10 0x00007f6faab4c8ec abort ./stdlib/abort.c:85:3
  #33 0x00007f6faab4e601 __libc_start_call_main ./csu/../sysdeps/nptl/libc_start_call_main.h:83:3
  #34 0x00007f6faab4e718 call_init ./csu/../csu/libc-start.c:128:20
  #35 0x00007f6faab4e718 __libc_start_main ./csu/../csu/libc-start.c:347:5
  #36 0x0000560c987f1465 (/jhbuild/install/bin/swift-frontend+0xd8a465)
  <unknown>:0: error: unable to execute command: Aborted
```

This was worked around with `-DCMAKE_Swift_COMPILATION_MODE=wholemodule`

I'm beyond my depth at the specifics of Swift-C++ compilation so I cannot say if this is an upstream bug or caused by the build.

Secondly:

```
: && /mnt/work/webkit/WebKit/Tools/Scripts/swift/swiftc-wrapper.sh -j 48 -num-threads 48 -emit-library -Onone -g  -Wl,--no-undefined  -fuse-ld=lld -Wl,--gc-sections -Wl,--disable-new-dtags --original-swift-compiler=/jhbuild/install/bin/swiftc @/mnt/work/webkit/WebKit/WebKitBuild/
  GTK/Debug/Source/WebKit/CMakeFiles/WebKit_SwiftInterop.dir/Debug/objects.rsp  -Xlinker -soname -Xlinker libwebkitgtk-6.0.so.4 -o lib/libwebkitgtk-6.0.so.4.18.0 @CMakeFiles/WebKit.rsp && :
  <unknown>:0: error: module name "webkitgtk-6.0.so.4.18" is not a valid identifier; use -module-name flag to specify an alternate name
```

This was solved by:

```diff
diff --git a/Source/WebKit/CMakeLists.txt b/Source/WebKit/CMakeLists.txt
index 2ae8e007fe70..3086fbd47bdc 100644
--- a/Source/WebKit/CMakeLists.txt
+++ b/Source/WebKit/CMakeLists.txt
@@ -1441,7 +1441,10 @@ if (DEFINED WebKit_SWIFT_INTEROP_SOURCES)
     set(_interop_objs "$<FILTER:$<TARGET_OBJECTS:WebKit_SwiftInterop>,EXCLUDE,\\.(g|p)ch$>")
     set(_interop_rsp "${CMAKE_CURRENT_BINARY_DIR}/CMakeFiles/WebKit_SwiftInterop.dir/$<CONFIG>/objects.rsp")
     file(GENERATE OUTPUT "${_interop_rsp}" CONTENT "$<JOIN:${_interop_objs},\n>")
-    target_link_options(WebKit PRIVATE "@${_interop_rsp}")
+    target_link_options(WebKit PRIVATE
+        "@${_interop_rsp}"
+        "$<$<LINK_LANGUAGE:Swift>:SHELL:-module-name WebKit>"
+    )
     set_property(TARGET WebKit APPEND PROPERTY LINK_DEPENDS "${_interop_objs};${_interop_rsp}")
 endif ()
```

This one just seems like a simple WebKit issue to resolve.

After this WebKit builds with Swift and the Swift code executes fine.

TingPing:
By building `swift-driver` WebKit builds out of the box without any patches.

nikolaszimmermann:
Wow massive amount of work! Will try to check the patch

TingPing:
Locally I have `swift-package-manager` and `sourcekit-lsp` built, so its truly a fully functional swift SDK, VSCode works, etc.

But it's painful and I've had enough pain. This is all we need for WebKit.

TingPing:
One new problem, `libLLVM.so` being shared actually conflicts with the host lib which Mesa pulls in. So I guess that can go back to static.

nikolaszimmermann:
> Locally I have `swift-package-manager` and `sourcekit-lsp` built, so its truly a fully functional swift SDK, VSCode works, etc.
> 
> But it's painful and I've had enough pain. This is all we need for WebKit.

As mentioned on Matrix, would be great to seperate concerns: produce a ephemeral 'builder' container that does the swift toolchain build, and then only copy the right artifacts into the final image..

````
FROM ubuntu:26.04 as builder
...do-the-build

FROM ubuntu:26.04
COPY --from=builder /all/we/need /target
```

So that the SDK package stays free from potential artifacts produced by the build...

PR #260, opened 2026-07-16
Cache downloads when building container image

TingPing:


nikolaszimmermann (commented):
These changes are fine per-se, I just worry about the disk usage on the runners. Shouldn't someone prune those directories from time-to-time at least to avoid unbounded growth? @clopez ?

TingPing:
There is a basic TTL `podman build --cache-ttl=1d` but we might want more control than that unfortunately.

clopez:
are downloads slow? which one specifically and how much it takes?

PR #261, opened 2026-07-16
wkdev-sdk: Maintain our own list of WebKit deps

TingPing:
While it's a clever idea to try and maintain one list of deps it simply doesn't reflect our usage here. It contains a large amount of things like the entire GStreamer stack and build only deps for the in-tree JHBuild modules. We don't need any of this.


TingPing:
The upstream list was also just really bad, I vaguely went through our actual cmake files to come up with a more useful list. Not confirmed to be perfect though.

nikolaszimmermann:
I think if we go that route, we should drop those install-*dep scripts in WebKit..

PR #266, opened 2026-08-12
wkdev-sdk: Move the WKDEV_SDK_VERSION ARG to preserve the build cache

clopez:
Podman makes each `ARG` available in the environment of every later `RUN` command. This meant that changing `WKDEV_SDK_VERSION` invalidated almost all cached layers, even though the version is only needed in the final label and version file.

Move the argument next to those uses at the bottom of the `Containerfile`.

This allows nightly and PR builds to share the expensive build layers.

This behavior can be reproduced with:

```
$ cat > Containerfile <<EOF
FROM ubuntu
ARG BUILD_MODE
RUN env | grep BUILD_MODE
EOF

$ podman build --build-arg BUILD_MODE=MagicValue01 .
[...]
STEP 3/3: RUN env | grep BUILD_MODE
BUILD_MODE=MagicValue01

$ podman build --build-arg BUILD_MODE=MagicValue02 .
[...]
STEP 3/3: RUN env | grep BUILD_MODE
BUILD_MODE=MagicValue02
```

See also:
https://docs.docker.com/reference/dockerfile/#impact-on-build-caching

Meanwhile at it remove also `ARG` `NUMBER_OF_PARALLEL_BUILDS` that was introduced on 9bcf627a but it is not not used anymore.

### DIFF
--- a/images/wkdev_sdk/Containerfile
+++ b/images/wkdev_sdk/Containerfile
@@ -3,17 +3,8 @@
 
 FROM public.ecr.aws/ubuntu/ubuntu:26.04
 
-# Single source of truth for the SDK version. The release workflow bumps the
-# -v<count> component and appends the source-repo -<gitsha>; everything else
-# (OCI label, /etc/wkdev-sdk-version, welcome message) reads from here.
-# The <major>.<minor> prefix tracks the upstream WebKitGTK / WPE WebKit release
-# branch and is bumped by hand here when moving to a new branch.
-ARG WKDEV_SDK_VERSION=2.53-v9-c666b76
-
-# Tweakable "make -j <x>" setting.
-ARG NUMBER_OF_PARALLEL_BUILDS=4
+# To change WKDEV_SDK_VERSION ARG variable check the bottom of this file.
 ARG CONTAINER_LOCALE=en_US.UTF-8
-
 # No need to modify these.
 ARG APT_UPDATE="apt-get update"
 ARG APT_BUILDDEP="apt-get --assume-yes build-dep"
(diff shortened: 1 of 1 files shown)


clopez:
Attached a test script to verify this with the current image
[test-podman-build-arg-cache.sh](https://github.com/user-attachments/files/30984187/test-podman-build-arg-cache.sh) (run it from the topdir of the repo)

Before this change:

```
/tmp/wkdev-build-arg-cache.XbSAhD $ grep 'elapsed' *.log
different-version.log:3644.25user 627.56system 15:25.94elapsed 461%CPU (0avgtext+0avgdata 2599324maxresident)k
original.log:3611.60user 615.37system 16:50.97elapsed 418%CPU (0avgtext+0avgdata 2648300maxresident)k
same-version.log:1.65user 1.35system 0:03.10elapsed 97%CPU (0avgtext+0avgdata 61772maxresident)k

/tmp/wkdev-build-arg-cache.XbSAhD $ grep -c 'Using cache' *
different-version.log:12
original.log:0
same-version.log:72
```

After this change:
```
/tmp/wkdev-build-arg-cache.LAadQq $ grep elapsed *.log
different-version.log:2.02user 1.25system 0:03.33elapsed 98%CPU (0avgtext+0avgdata 66536maxresident)k
original.log:3613.76user 622.00system 15:16.61elapsed 462%CPU (0avgtext+0avgdata 2666968maxresident)k
same-version.log:1.78user 1.25system 0:03.18elapsed 95%CPU (0avgtext+0avgdata 62268maxresident)k

/tmp/wkdev-build-arg-cache.LAadQq $ grep -c 'Using cache' *
different-version.log:71
original.log:0
same-version.log:72
```

nikolaszimmermann (approved):
Wonderful - I was under the impression we already fixed that, wrong I was 😄

PR #268, opened 2026-08-13
CI: Add daily podman storage maintenance for self-hosted runners

clopez:
The CI jobs already clean up after themselves at the end of each run, but that isn't enough to keep the disk under control: when a build is aborted mid-way (for example, a new push to a PR cancels the build that was already running), it dies before its cleanup steps run and leaves behind half-built images and buildah containers that nothing ever removes. Over time the runners fill up anyway.

This adds a workflow that runs daily (plus manually, with an optional force=true input) on every runner. If the runner has less than 50GiB free, or if the reset is forced, it wipes the podman storage with `podman system reset` and immediately rebuilds the wkdev-sdk image to warm the build cache.
Otherwise it just records the disk usage stats, which over time gives us a daily record of how fast each disk fills. A reset is used instead of a prune because it also recovers from corrupted storage and removes buildah leftovers without relying on podman bookkeeping being intact.

Since GitHub Actions can't target "all runners with a label", the matrix lists every machine by a unique label equal to its runner name. The build and deploy jobs check their own runner name against that list and fail loudly if it's missing, so a forgotten runner shows up as a red check instead of silently never being cleaned.

The maintenance jobs also verify that they landed on the machine each matrix entry names, catching a unique label assigned to the wrong runner before the wrong machine is reset. Concurrency is scoped per runner, so an unavailable machine does not delay maintenance of the healthy ones.

With disk usage now under control, the build job can also stop deleting the image it just built. Deleting it threw away the cache that makes the next push to the same PR build much faster. Instead, only the older images of the same PR are removed. Images from closed PRs and other leftovers are left for the maintenance workflow. Release builds keep their image too: it's the best possible cache for the PR builds that follow.

Older PR images are only removed when the current image exists locally, so a failed build cannot discard the last useful cache. The deploy job remembers which images were already present locally, so it cleans only its own artifacts to ensure that it doesn't discard a build cache that it did not create. The cleanup runs with "if: always()", so architecture images newly pulled to assemble the manifest are reclaimed even when the manifest push fails.

The helper scripts live in .github/scripts: disk-usage.sh owns the 50GiB threshold and the before/after disk reports (shown in the job summary and kept as a 90-day artifact), build-wkdev-sdk-image.sh is shared by the real build and the cache-warming rebuild so both always build the same way, and assert-runner-in-maintenance-matrix.sh is the runner coverage check shared by the build and deploy jobs. They are careful about degraded machines: a broken podman is reported as such and treated as "maintenance needed", missing or malformed snapshots skip the comparison instead of inventing numbers, unavailable optional measurements are reported as "not measured", and no failure can leave a partial or stale report behind.

The helper scripts have regression tests that stub podman and the disk tools, so their success, failure, and degraded-machine paths run in seconds.

### DIFF
--- a/.github/scripts/assert-runner-in-maintenance-matrix.sh
+++ b/.github/scripts/assert-runner-in-maintenance-matrix.sh
@@ -1,0 +1,77 @@
+#!/bin/bash
+# Fails (exit 1) when the runner executing this job is not listed in the
+# podman-storage-maintenance.yaml matrix. The matrix entries are runner
+# names (used as unique labels), so a runner missing from that list never
+# gets its storage reset and would silently fill up its disk.
+#
+# Expects RUNNER_NAME in the environment (the workflow passes runner.name).
+# The match is an exact whole-string comparison against every `runner:` value
+# of the matrix, so a runner name is never treated as a pattern (no false
+# match) and quoted entries, names containing spaces or a trailing comment on
+# the matrix line are all handled (no false "not covered" failure).
+#
+# Exits 2 when the matrix itself cannot be read: no entries at all, a
+# flow-style entry ({arch: ..., runner: ...}), or an entry this script
+# cannot parse (unclosed quote, a comment-like '#' inside a quoted name, or
+# a quote character anywhere in a name). Those get their own error, because
+# "I cannot tell" and "this runner is not covered" call for different fixes.
+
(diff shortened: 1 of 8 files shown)


nikolaszimmermann (approved):
rs=me

PR #275, opened 2026-08-28
CI: Show the disk usage on the run page instead of only inside an artifact

clopez:
The maintenance workflow collected disk stats every day but only wrote a summary on the rare days it actually reset a runner. A normal run said just "Free disk space is above the threshold: reset skipped".

But the numbers about how much free disk space is available and how much is used were available only by downloading and unzipping the snapshot artifact, which is far from convenient.

This patch adds a new "summary" action to disk-usage.sh that generates a small table with the storage details in markdown, so every run calls it (even when no reset happens), so when opening the run it will show all the details of the storage for the available runners. When a reset happens, then the same table is rendered again from the final snapshot, and the comparison of the two follows it: before, after, and what changed.

This patch also separates report generation from GitHub-specific output handling. disk-usage.sh now writes Markdown reports to stdout and diagnostics to stderr. The workflow decides where those reports go, using `tee -a "$GITHUB_STEP_SUMMARY"` to show the same output in both the step log and the workflow run summary.

Previously, "compare" printed one report to the log and independently appended another to "$GITHUB_STEP_SUMMARY". It now produces a single Markdown report that is sent to both the job summary and the step log, and accepts a heading to identify each section.

The artifacts keep their 90 day retention: they are now the archive rather than the only way to read a run.


nikolaszimmermann (approved):
Nice, r=me

PR #276, opened 2026-08-28
Remove armhf (32-bit ARM) cross-compilation support

clopez:
The armhf cross toolchain shipped in the SDK image is legacy and no longer supported, so drop it:

  - Delete images/wkdev_sdk/cross/armhf/ (the gcc/g++-arm-linux-gnueabihf toolchain list, the libicu-dev:armhf JSCOnly dependency list and the GCC/Clang CMake toolchain files).

  - Containerfile: drop the "COPY /cross .", the generation of ubuntu-armhf.sources pointing at ports.ubuntu.com along with the Architectures: pinning of ubuntu.sources and "dpkg --add-architecture armhf" (all of which only existed to serve armhf packages), the LST_FILES conditional that pulled in armhf/*.lst, and the step installing the toolchain files into /usr/share/wkdev-sdk/cross/armhf/.

  - wkdev-create: make the try_process_nvidia_gpu call unconditional again, reverting a807b60 ("Skip nvidia configuration when --arch is supplied."), which only existed to fix 32-bit containers.


nikolaszimmermann (approved):
Yes!

